PR #4, opened 2019-08-05
Cleanup of documentation

bjhandeland:
Cleaned up the formatting and spelling of the readme file, and went through it all being as meticulous as possible as we'll have to use it extensively moving forwards with the new client. 

I made some changes, but I have some questions about additional things I would like to change:

Typos:

1. There's something called `alarm_applience`. Is this a typo in the docs or a typo in the implementation? Can this be changed to `alarm_appliance`? If yes, good. If no, we should add a comment to the docs about the typo and pref. work towards fixing it.

2. There's a supported status named `activ` and `deactiv`. Is this a typo? Can this be changed to `active` and `deactive`? If yes, good. If no, we should add a comment to the docs about the typo and pref. work towards fixing it.

3. There's an event called `spinnning`. Is this a typo? Can this be changed to `spinning`? If yes, good. If no, we should add a comment to the docs about the typo and pref. work towards fixing it.

4. There's a unit called `galon`. Is this a typo? Can this be changed to `gallon`? If yes, good. If no, we should add a comment to the docs about the typo and pref. work towards fixing it.

Other: 

1. There are quite a few sensors services, but for some reason. `sensor_contact` and `sensor_presence` are listed separately. Why is this? Can they be merged with the rest of the sensors? If the reason is that they report value of bool instead of float, can't we create a category for binary sensors, just like we have numeric sensors?

2. For some services, props are listed, but it seems like only some of them are listed. For example `out_lvl_switch`: there are three props listed, but there are also properties associated with the interfaces themselves? Is there perhaps a better way this can be structured?

3. The value example for the `sw_type` props for `out_lvl_switch` service is a bit confusing. I'm assuming it should be just one string or can it be an array of strings?

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Service name  | Units                                     | Description
 ------------- |-------------------------------------------|------------
 `meter_elec`  | kWh, kVAh, W, pulse_c, V, A, power_factor | Electric meter
 `meter_gas`   | cub_m, cub_f, pulse_c                     | Gas meter
-`meter_water` | cub_m, cub_f, galon, pulse_c              | Water meter
+`meter_water` | cub_m, cub_f, gallon, pulse_c              | Water meter
 
 #### Interfaces
 
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ Service name        | Event                                   | Description
 `alarm_system`      | hw_failure, sw_failure, hw_failure_with_code, sw_failure_with_code |
 `alarm_emergency`   | police, fire, medical                   |
 `alarm_time`        | wakeup, timer_ended, time_remaining     |
-`alarm_applience`   | program_started, program_inprogress, program_completed, replace_filter, set_temp_error, supplying_water, water_supply_err, boiling, boiling_err, washing, washing_err, rinsing, rinsing_err, draining, draining_err, spinnning, spinning_err, drying, drying_err, fan_err, compressor_err |
+`alarm_appliance`   | program_started, program_inprogress, program_completed, replace_filter, set_temp_error, supplying_water, water_supply_err, boiling, boiling_err, washing, washing_err, rinsing, rinsing_err, draining, draining_err, spinnning, spinning_err, drying, drying_err, fan_err, compressor_err |
 `alarm_health`      | leaving_bed, sitting_on_bed, lying_on_bed, alarm_health, posture_change, sitting_on_bed_edge, alarm_health, volatile_organic_compound |
 `alarm_siren`       | inactive, siren_active                  |
 `alarm_water_valve` | valve_op, master_valve_op, alarm_water_valve, valve_short_circuit, current_alarm, alarm_water_valve, master_valve_current_alarm |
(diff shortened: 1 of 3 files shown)


chribru:
Ultimately it's up to Aleks to decide and know these things, but I'll try to answer what I can, I what I think :)

Typos:
1, 3 and 4 are definitely typos. None of them are integrated anywhere, but might be in use by some Z-Wave devices. none of the ones we officially support, though, as far as I know. In my opinion these should be corrected before they are used any further.

Regarding 2 (activ/deactiv) This is short for activated and deactivated, as far as I know. In my opinion, it would be quite risky to change it, as it's used by adapters, core, kind owl, and possibly also external alarm system integrations. But a note in the documentation about the spelling is a very good idea.

Other:
1. It looks to me too as it is that way because they are binary sensors. I think the idea was that the general sensor was float, and then there are a few exceptions of other types. Right now we have 2 sensors which are binary as well, but there might be other types in the future? 

Not sure if we want to make a system out of it or how it would look. Maybe it could just be evt.sensor.report for all of them, and then the type of the interface would tell what kind of sensor it was. But generally, one interface should have a defined type, and evt.sensor.report is defined to be float, so there would have to be different interfaces for different types.

2. As I understand it, props listed under the service are extra info about the service, such as what kind of multilevel switch it is. They are part of the site shadow. The props under commands are more like extra arguments to the command, some of which are optional.

I am fairly certain that all props are strings, which is also why we have  {"duration": "5"} and not {"duration": 5}. And the same for events. For example delta_t for meter_ services. One could argue that is should be part of the val, though.

3. Since props are strings, I would say only one of them. It also makes sense logically in this case - it's basically either dimmer or roller shutter in practice for those examples.

bjhandeland:
Thank you for the detailed feedback @chribru!

Based on your comments I did the following:

1. Fixed the typos, with the exception of activ and deactiv where I added a comment instead.

2. I grouped the two binary sensors under "Binary Sensors" and made a list of supported sensor services that are binary sensors instead. Documentation remains the same, just the grouping is a little different.

3. I split props into "Interface props" and "Service props" to indicate that there's a difference. This also frees up space to add descriptions about the actual interface, not just the props for the interface in the tables.

I think we'll have to wait for @alivinco to review the rest before we merge?

chribru:
Yes, let's wait for @alivinco to review the rest :+1:

bjhandeland:
I sorted interfaces by their attributes and sorted all lists alphabetically where possible. Now consistency all interfaces will show inputs first, then outputs, each sorted by their respective attribute.

One thing I came across that you might be able to answer @chribru:

For the `barrier_ctrl` service, it has the attribute `state` as well as `tstate`. This looks suspiciously much like a typo where the `tstate` attribute should in reality be `state.

For the same service, is `notiftype` a typo that should be `notifytype`? Seems silly to abbreviate just one letter. I would suggest adding a note in the case that it's actually `notiftype`.

chribru:
Regarding `notiftype`, I think it's intentional.

bjhandeland:
Updated the docs based on talk we had yesterday @chribru. I changed the intro to cover everything about services, interfaces, attributes and topics to the best of my knowledge. I think this completes all the changes I would like to make for now, and we'll wait for @alivinco to review.

bjhandeland:
@alivinco have you had a chance to look at this?

bjhandeland:
@alivinco

PR #7, opened 2019-09-21
Added headers to fimp message format.

bjhandeland:
Makes links to docs better.


bjhandeland:
@chribru , @alivinco any update here?

PR #13, opened 2019-11-19
link was broken

maverich:


maverich:
https://github.com/futurehomeno/fimp-api#adding--removing-things-to-fh-system
#1's link is broken.

SuperKogito:
I think broken links are a common issue and a radical solution to this problem is automatically checking and reporting them. Luckily the new GitHub actions and  workflows help with that. I recently wrote an action that just does that. [URLs-checker](https://github.com/marketplace/actions/urls-checker) will check and report broken links, which is useful to spot similar issues to this one.

PR #25, opened 2020-07-14
Added cmd.alarm.clear to alarm service

tskaard:
Added interface cmd.alarm.clear to alarm services


tskaard:
@ditek no I mean device, or to be specific I mean service for a device.
But the adapter can generate the message on behalf of the device 😃

ditek:
I see. Then all is good.

PR #35, opened 2020-11-10
Updated dev_sys wit blocking, wake up timer and reinterview commands

imilicevic7:
Updated API with new command (blocking feature) and added some old-never-documented commands.


bjhandeland (approved):
LGTM

PR #41, opened 2022-04-12
#noisue add new alarm services based on zwave notificaitons v8

eternalamenity:


eternalamenity:
This is one of proposed solutions, which introduces a new way of understanding "Notifications" from a Zwave perspective.
Another solution is to create a separate service called eg. "Notification", which would not affect actual implementation of alarms but might cause different issues.

PR #43, opened 2022-05-04
#noissue update barrier_ctrl documentation

eternalamenity:


README.md
```diff
@@ -878,7 +878,7 @@ out  | evt.notiftype.report     | bool_map   |
 in   | cmd.op.stop              | null       | Emergency stop of any operation.
 -|||
 in   | cmd.state.get_report     | null       | Get current state
-out  | evt.state.report         | string     | Current state
+out  | evt.state.report         | string     | Current state (one of sup_states or number)
```
bjhandeland, line 881: Can you provide more information on why / when it would be a number? Also what does the number represent?

bjhandeland (changes requested):
See comment.

README.md
boguslaw-wojcik: Actually, it is heavily against any consistency to have "state report" report anything else than one of the supported states. If there is something like progress needed it should have a deisgnated interface? What is the context of it?

eternalamenity: For Barrier Operator in zwave, a device can report with such values:
Screenshot from 2022-05-04 17-43-59


also controlling zwave specification of this functionality says what is last known barrier state:
"Last known Barrier State (Open, Closed, stopped at a % position or unknown)" 

And also to be honest, this is only documentation thing. In zwave those changes exist for a long long time (not introduced by me).

eternalamenity: @bjhandeland @boguslaw-wojcik

boguslaw-wojcik: As discussed on Slack:

We should not mix enums with dynamic values. States are enums, while the position of the barrier is a dynamic value. We should either interpret 0x01..0x63 values just as "Stopped" and optionally introduce another interface for getting the current position.

Please do confirm with @chribru and investigate if this won't create any backwards compatibility breaks.

eternalamenity: https://github.com/futurehomeno/fimp-api/pull/45

PR #49, opened 2022-05-26
SDU-1875 introduce zwave specific service

eternalamenity:
If someone wants to know more about why this service is being created, PM me or write a comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1244,8 +1244,9 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
 }
 ```
 
-### Zwave Specific service 
-This service is used to meet all zwave specific requirements that are not satisfied with already implemented services.
+### Notification service
+
+This service is used to receive notifications from devices.
 
 #### Service name 
 
@@ -1255,7 +1256,7 @@ This service is used to meet all zwave specific requirements that are not satisf
 
 Type | Interface               | Value type | Description 
 -----|-------------------------|------------|--------------
-out  | evt.notification.report | object     | Notification reports that are sent by slave devices.
+out  | evt.notification.report | object     | Notification reports that are sent by devices.
 
 #### Interface storage
 
--- a/README.md
+++ b/README.md
@@ -1249,19 +1249,20 @@ This service is used to meet all zwave specific requirements that are not satisf
 
 #### Service name 
 
-`zwave_specific`
+`notification`
 
 #### Interfaces
 
 Type | Interface               | Value type | Description 
 -----|-------------------------|------------|--------------
-out  | evt.notification.report | str_map    | Notification Reports that are sent by slave devices.
+out  | evt.notification.report | object     | Notification reports that are sent by slave devices.
 
-#### Storage property
+#### Interface storage
 
-Key        | Value description 
------------|----------------------------------
-sub_value  | Values pair to store in Vinculum.
+Name        | Value example                 | Description
+------------|-------------------------------|-------------
+`sub_value` | "home_security:sensor_status" | With usage of sub_value, Vinculum will know that it has to store separate value for every pair of choosen properties for given notification.
+
 
 #### Examples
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1243,3 +1243,78 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+#### Service name 
+
+`zwave_specific`
+
+#### Interfaces
+
+Type | Interface               | Value type | Description 
+-----|-------------------------|------------|--------------
+out  | evt.notification.report | str_map    | Notification Reports that are sent by slave devices.
```
bjhandeland, line 1258: nit: change to `Notification reports that are sent by worker devices.` 

I changed slave to worker, but if slave is used in the zwave documentation, we should stick with that.

README.md
```diff
@@ -1243,3 +1243,78 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+#### Storage property
+
+Key        | Value description 
+-----------|----------------------------------
+sub_value  | Values pair to store in Vinculum.
```
bjhandeland, line 1264: I don't get what `sub_value` is? Subscription value? Typo for supported value? And I don't get what the description means either :|

README.md
```diff
@@ -1243,3 +1243,78 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+#### Examples
+
+Motion Detection occurs for state Motion Sensor Status:
```
bjhandeland, line 1268: I would use the value names in the header here to make it easier to read the example, e.g.

"`motion_detection` detected for state `motion_sensor_status`.".

README.md
```diff
@@ -1243,3 +1243,78 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+```json
+{
+   "type": "evt.notification.report",
+   "serv": "zwave_specific",
+   "val_t": "str_map",
+   "val": {
+      "notificationEventState": "motion_detection",
+      "notificationType": "home_security",
+      "stateVariable": "motion_sensor_status"
```
bjhandeland, line 1278: If these are zwave specific key names, can you please dcoument what they represent? It's not immediately clear to me what they are.

If they are not zwave specific and they are keys we have added, can we revise the names?

notificationType: type of the notification, gotcha.
notificationEventState: why not just notificationEvent?
stateVariable: why not just notificationState

Either way, these are not that clear and should be documented / revised.

README.md
```diff
@@ -1243,3 +1243,78 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+   },
+   "storage": {
+     "sub_value": "home_security:motion_sensor_status"
+   }
+}
+```
+
+Motion Sensor Status state has been changed to State Idle:
```
bjhandeland, line 1286: Change title to use value names for clarity.

bjhandeland (changes requested):
see comments

README.md
eternalamenity: It defines how values are stored in Vinculum persistant storage. It defines a pair of keys which should be treated as unique to hold as persistant data. 
For example now for every evt.notification.report Vinculum would keep only one value.
But I want a Vinculum to hold every value with unique pair of keys eg. "home_security:motion_sensor_status".
So in that way when another notification occurs (eg. home_security:cover_status), Vinculum will not replace previous one, but will keep both.

README.md
eternalamenity: From zwave perspective "slave" means a device that is added to a so-called "controller" and cannot operate itself with network's capabilities. So I think we can stick with that.

bjhandeland: Sounds good. Still change it to `Notification reports that are sent by slave devices.` :)

bjhandeland (approved):
lgtm

README.md
eternalamenity: fixed, use proper description now

bjhandeland (approved):
lgtm

PR #50, opened 2022-05-30
SDU-1850 support for start fading/enhancing and stop fading/enancing …

eternalamenity:
Functionality required by specification

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,6 +762,8 @@ Type | Interface            | Value type |  Description
 -----|----------------------|------------|-------------------
 in   | cmd.color.get_report | null       | The command is a request for a map of color component values
 in   | cmd.color.set        | int_map    | value is a map of color components. val= {"red":200, "green":100, "blue":45}
+in   | cmd.color.fade_enhance | str_map  | Value is a map of color:direction. val= {"red": "up"}, val = {"green": "down"}
+in   | cmd.color.stop       | str        | Stop fading/enhancing single color component. val = "red"
 out  | evt.color.report     | int_map    | Map of color components, where value is component intensity.
 
 #### Service props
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -762,6 +762,8 @@ Type | Interface            | Value type |  Description
 -----|----------------------|------------|-------------------
 in   | cmd.color.get_report | null       | The command is a request for a map of color component values
 in   | cmd.color.set        | int_map    | value is a map of color components. val= {"red":200, "green":100, "blue":45}
+in   | cmd.color.fade_enhance | str_map  | Value is a map of color:direction. val= {"red": "up"}, val = {"green": "down"}
```
boguslaw-wojcik, line 765: Let me spare you pain from Bjarne. You need to align columns.

boguslaw-wojcik (changes requested):
Some suggestions and style fix.

README.md
eternalamenity: 2. I can make it one example, but this action should actually only take ONE color component and it's direction. So it will be val={"red": "up"}
3. Yeah, I can change that
4. Isn't that every interface is optional by default? As far as I've been talking with @chribru , he said me that a few times.
5. This action is 1-1 the same action that zwave represents. I don't know what is better, it's either sending X fimp commands from mobile (if user wants to stop transition change for a few colors) or having loop in zwave adapter. Also I don't know if we will meet below mentioned require
Screenshot from 2022-05-31 12-59-10
ment:
Screenshot from 2022-05-31 12-58-54

boguslaw-wojcik: Example object: 
{
"component": "blue",
"transition": "down"
}

eternalamenity: fixed

boguslaw-wojcik (approved):

PR #52, opened 2022-06-07
SDU-1850 add duration property for start transition color service

eternalamenity:


README.md
```diff
@@ -789,6 +790,8 @@ Supported `temp` values: 1-65279 mired. Actual color temperature supported by en
 
 - Mix of warm white intensity and cold white intensity forms color temperature.
 
+- Duration property within cmd.color.start_transition command is optional and means seconds.
```
GenixPL, line 793: it should be mentioned that above `maxSeconds` seconds will not be handled and the value will be cut into full minutes, with example, sending 128 (seconds, == 2 min 8 seconds) will result in adapter setting only 2 min and not 2m 8s

GenixPL (changes requested):

README.md
eternalamenity: Extended description as much as I can :D 
If someone is able to describe it better, then write it down please so I could update it.

README.md
```diff
@@ -764,14 +764,15 @@ in   | cmd.color.get_report       | null       | The command is a request for a
...
 
 #### Service props
 
-Name             | Value example            | Description
------------------|--------------------------|-------------
-`sup_components` | ["red", "green", "blue"] | List of supported color components
+Name             | Value example                                                            | Description
+-----------------|--------------------------------------------------------------------------|-------------
+`sup_components` | ["red", "green", "blue"]                                                 | List of supported color components
+`duration`       | {"minSeconds": 1, "maxSeconds": 127, "minMinutes": 1, "maxMinutes": 127} | Supported duration steps for transition change (optional).
```
boguslaw-wojcik, line 775: Wouldn't it be easier to just mention: min: 1, max: 7620? The rounding up to one minute after 127 seems adapter specific and not that important...

GenixPL: the idea was to be as specific as possible in here so that the app _can_ provide proper ui that changes the step when needed. 
I think it's better to optionally lose that information in ui than in service because if service doesnt provide it we have no way of knowing about it in ui.
the worst cause I believe would be when an adapter wouldn't support seconds at all, then setting sth small would result in either having 0 or 1 min which would be quite significant

bjhandeland: This seems a bit odd

The `start_transition` supports passing a duration now, which in the example is `127`.
The `duration` value example seems to indicate that this can be an object with a whole bunch of things.

Question is then, what is the `duration` in the `start_transition`?
Would it not be better to have the duration in min max milliseconds and then have a "steps" property?

bjhandeland: Also, should this not be `sup_durations`?

README.md
```diff
@@ -789,6 +790,12 @@ Supported `temp` values: 1-65279 mired. Actual color temperature supported by en
 
 - Mix of warm white intensity and cold white intensity forms color temperature.
 
+- Duration property within cmd.color.start_transition command is optional and means seconds.
+
+- For Z-Wave protocol, every duration value above the 'maxSeconds' value will not be handled and will be cut into full minutes.
+  Example: sending "duration": 128 (seconds = 2 minutes and 8 seconds) will result in adapter setting only 2 minutes within device.
+  Duration increment step should be '1' for seconds: 1, 2, 3...125, 126, 127 and '60' for minutes, starting from: 180, 240, 300...7620.
```
bjhandeland, line 797: This seems extreme. I would prefer a solution that's a bit more straight forward.

bjhandeland (changes requested):
See comments

README.md
eternalamenity: @bjhandeland You're right. Duration in start_transition interface are seconds. 
Here it should be named "sup_durations" instead of just "durations"! 
Fixing

README.md
boguslaw-wojcik: More like steps:
sup_durations: [
{min: 1, max: 127, step: 1},
{min: 180, max: 7620, step: 60}
]

bjhandeland: Alternatively we can just have 

`sup_durations: {min: 1000, max: 999999}`

Where the duration would always be in milliseconds. I don't think the protocol should worry about how things are to be displayed. Is there a technical reason why it's split into seconds and minutes?

eternalamenity: @boguslaw-wojcik seems nice, thank You, will fix it

eternalamenity: fixed

README.md
```diff
@@ -764,14 +764,15 @@ in   | cmd.color.get_report       | null       | The command is a request for a
...
 
 #### Service props
 
-Name             | Value example            | Description
------------------|--------------------------|-------------
-`sup_components` | ["red", "green", "blue"] | List of supported color components
+Name             | Value example                                                              | Description
+-----------------|----------------------------------------------------------------------------|-------------
+`sup_components` | ["red", "green", "blue"]                                                   | List of supported color components
+`sup_durations`  | [{"min": 1, "max": 127, "step": 1}, {"min": 180, "max": 7620, "step": 60}} | Supported duration steps for transition change (optional). Min and max always reflects seconds.
```
boguslaw-wojcik, line 775: Array should end with }] not }}

README.md
```diff
@@ -789,6 +790,10 @@ Supported `temp` values: 1-65279 mired. Actual color temperature supported by en
 
 - Mix of warm white intensity and cold white intensity forms color temperature.
 
+- Duration property within cmd.color.start_transition command is optional and means seconds.
+
+- For Z-Wave protocol, slave devices will accept only values from range 1...127 seconds or 1...127 minutes. This is why sup_durations property was introduced. Zwave adapter will be responsible for converting seconds above value 127 to minutes.
```
boguslaw-wojcik, line 795: MAYBE more formal:

- In the Z-Wave protocol color component transition duration can only be set between 1 - 127 seconds or 1 - 127 minutes. In order to fully support the Z-Wave protocol, sup_durations property can contain an array of ranges.

boguslaw-wojcik (changes requested):
Array typo.

README.md
eternalamenity: fixed

README.md
eternalamenity: done

boguslaw-wojcik (approved):

GenixPL (approved):
git

bjhandeland (approved):
lgtm

PR #53, opened 2022-06-08
SDU-1861 documentation for sub_value property within 'storage' proper…

eternalamenity:
The idea figured out right with Christer was to differentiate that information from "props", because it fulfills completelly different role in an mqtt messege. 
This information is not a typical property. It says how to persistently store message's val in Vinculum's State Storage.


boguslaw-wojcik (commented):
1) Can this storage be multi-level? To keep the values, in this example per cool and unit C? While it does not make sense in this example, it might in future.

2) Prolly you should also add it to the water heater and meter reports for units (and all currently supported multi-value reports).

boguslaw-wojcik (approved):

PR #54, opened 2022-06-13
SDU-1854 new meter services, props and storage

eternalamenity:


README.md
```diff
@@ -321,11 +324,19 @@ in   | cmd.meter_ext.get_report      | null       |                         | Re
...
------------|---------------|-------------
-`delta_t`  |               | time delta
-`prv_data` |               | previous meter reading
-`unit`     |               |
+Name         | Value example | Description
+-------------|---------------|-------------
+`delta_t`    |               | time delta
+`prv_data`   |               | previous meter reading
+`unit`       |               | One of sup_units. For meter_unknown it will be a number.
+`rate_type`  | "export"      | Defines whether reading is based on consumption (import) or production (export).
```
boguslaw-wojcik, line 332: "direction" seems to be better and more consistent with extended meter properties.

eternalamenity: More consistent with extended meter properties? I did not see it anywhere there(?)
Imo I see "rate_type" as exactly what it represents. Direction would match more to values like "up"/"down"/"left"/"right" to me.

README.md
```diff
@@ -321,11 +324,19 @@ in   | cmd.meter_ext.get_report      | null       |                         | Re
...
 #### Interface props
 
-Name       | Value example | Description
------------|---------------|-------------
-`delta_t`  |               | time delta
-`prv_data` |               | previous meter reading
-`unit`     |               |
+Name         | Value example | Description
+-------------|---------------|-------------
+`delta_t`    |               | time delta
```
bjhandeland, line 329: Since we're fixing this, can we add a bit more descriptive info here. Maybe `Time (in seconds) since the last reading.` I'm not sure it's in seconds, so change to whatever is appropriate.

README.md
```diff
@@ -321,11 +324,19 @@ in   | cmd.meter_ext.get_report      | null       |                         | Re
...
+`prv_data`   |               | previous meter reading
```
bjhandeland, line 330: Change to `Previous meter reading.`.

README.md
bjhandeland: Agree with @eternalamenity here from a semantic point of view.

If the functionality is indeed different and we should support words like "up", "down", etc. only then should we change the key.

README.md
```diff
@@ -303,11 +303,14 @@ Meters report consumption over the service.
...
-`meter_gas`   | cub_m, cub_f, pulse_c                     | Gas meter
-`meter_water` | cub_m, cub_f, gallon, pulse_c             | Water meter
+Service name    | Units                                     | Description
+----------------|-------------------------------------------|------------
+`meter_elec`    | kWh, kVAh, W, pulse_c, V, A, power_factor | Electric meter
+`meter_gas`     | cub_m, cub_f, pulse_c                     | Gas meter
+`meter_water`   | cub_m, cub_f, gallon, pulse_c             | Water meter
+`meter_heating` | kWh                                       | Heating meter
+`meter_cooling` | kWh                                       | Cooling meter
+`meter_unknown` | Numbers representing units.               | Unknown meter
```
bjhandeland, line 313: nit: add a `.` after the description.

bjhandeland (changes requested):
see comments

README.md
```diff
@@ -321,11 +324,19 @@ in   | cmd.meter_ext.get_report      | null       |                         | Re
...
-`delta_t`  |               | time delta
-`prv_data` |               | previous meter reading
-`unit`     |               |
+Name         | Value example | Description
+-------------|---------------|-------------
+`delta_t`    |               | time delta
+`prv_data`   |               | previous meter reading
+`unit`       |               | One of sup_units. For meter_unknown it will be a number.
+`rate_type`  | "export"      | Defines whether reading is based on consumption (import) or production (export).
+`meter_type` | 10            | Number representing meter type. Applicable ONLY for meter_unknown service.
```
bjhandeland, line 333: Do you have any more information on what meter types this would be?

bjhandeland (changes requested):
one more

README.md
boguslaw-wojcik: 1) Please see this graph: https://github.com/futurehomeno/fimp-api/blob/master/static/inverter.png
2) Here we have extended values of meter elec: https://github.com/futurehomeno/fimp-api#extended-report-object
3) I thought it was finalized, however, it seems it stayed just internal in ecollector and was never put into spec. In order to cope with single meter report having no way to express whether it is imported or exported energy there was a proposal for additional property "dir(ection)".

Now, lexically, rate is "a measure, quantity, or frequency, typically one measured against another quantity or measure" or "assigned a standard or value to (something) according to a particular scale".

While direction is "a course along which someone or something moves."

There are definitely better words to describe state of being exported or imported, but the direction is unequivocally better than rate for this purpose.

boguslaw-wojcik: Now the big problem - we cannot implement it, because all current usages assume that electricity meter reports are of import type. I have no immediate solution to this, but this is a BLOCKER.

boguslaw-wojcik: Ok, after quick talk I think we can go with following implementation:

1) For meter_elec service use ONLY extended meter report to report export values. Support simple reports for import values as usuall.
2) For any other meter include this direction/flow/etc just not "rate" property with enum values import/export. Put in description that if not present import or N/A should be inferred.
3) Add a big warning somewhere: For backwards compatibility, meter_elec simple reports should only support "import" direction.
4) I think we should add service property "sup_directions" to indicate whether this service supports it or not.
5) This property then would have to be added and supported on command as well to request either import or export. 

This seems to be more complicated than you think :P

eternalamenity: Applied documentation fixes according to talks with @chribru and @boguslaw-wojcik.
Handling unknown meter values will be done in a separate Pull Request.

README.md
eternalamenity: changed

README.md
eternalamenity: deleted this property, because it will be handled using separate service within separate Pull Request :)

bjhandeland (approved):
lgtm

PR #56, opened 2022-06-22
Fixed level switch specification.

boguslaw-wojcik:


README.md
```diff
@@ -271,29 +271,29 @@ Used for dimmers and things generally controlled with sliders.
 
 Type | Interface          | Value type | Properties              | Description
 -----|--------------------|------------|-------------------------|------------
-in   | cmd.binary.set     | bool       |                         | true is mapped to 255, false to 0
-out  | evt.binary.report  | bool       |                         |
+in   | cmd.binary.set     | bool       |                         | True is mapped to 255, false to 0.
+out  | evt.binary.report  | bool       |                         | Deprecated. Service should use `evt.lvl.report` instead.
```
eternalamenity, line 275: As far as I've tested it... evt.binary.report shows a status of an on/off button, and evt.lvl.report will show level using slider.
Both are independent so If we would like to get rid of evt.binary.report and leave it as just for backward compatibility, then it would need mobile app to have changes and show button status based on evt.lvl.report. I think also cmd.binary.set i send from mobile when using button.

chribru (dismissed):
LGTM

README.md
chribru: as I understand the app already supports using the value from `evt.lvl.report` if no `evt.binary.report` has been sent. Having both is causing some issues, because they're kinda not independent. For example if a lvl is reported after a binary false, it means that the binary false is not valid anymore, right?

chribru: Btw, we are not talking about removing cmd.binary.set, just that it should result in an evt.lvl.report instead of an evt.binary.report

eternalamenity: Screenshot from 2022-06-22 14-08-06
If mobile is already implementing it then fine. But also you have to know that we have to give possibility to "Restore most recent (non-zero) level". Here is the requirement:

Screenshot from 2022-06-22 14-09-42

chribru: yes, we should, and that can still be done with `cmd.binary.set`, just that after setting the value, the adapter would report the level with `evt.lvl.report` instead of sending an `evt.binary.report`. What would a device typically report on Z-Wave when you use the 0xFF value, btw? Would it give you the value, or is that something the adapter would have to poll to be able to report?

chribru: IMO would should also remove the "true will be mapped to 255" and forbid using the special 255 value for setting on, instead `cmd.binary.set` should be used instead. That way it won't interfere with dimmers whose range exceeds 255

eternalamenity: Dimmers range exceeds 255? Not in zwave i think.

chribru: No, I was thinking in general  :slightly_smiling_face: Because having this Z-Wave specific 255 value in the FIMP API could come in the way of dimmers going above 255 if a value in the middle of the range needs to mean something special. Instead `zwave-ad` could reject any `cmd.lvl.set` above 100 and require a `cmd.binary.set` to send the 0xFF value on Z-Wave?

eternalamenity: Yeah that could be done. cmd.binary.set with value true - might be interpreted like "Restore to last known non zero value", sure.

eternalamenity: So @boguslaw-wojcik , If You will create tasks for me and will mention it to the mobile team. Then go ahead and resolve this convertasion :)

eternalamenity (approved):

README.md
boguslaw-wojcik: @eternalamenity @chribru Changed wording to: "True should result in setting last known non-zero level or fallback to maximum."

chribru: should the fallback be to maximum or could it be another reasonable default? I think Vinculum uses 75% or so for EnOcean dimmers if it doesn't know the previous value.

chribru (approved):
still LGTM

eternalamenity (approved):

PR #59, opened 2022-07-17
Generic changes, tables formats

BudzikM:
I cannot add reference to the new service... let me know how to do it.

Reviewed and propsed changed to elec_meter, added some descriptions and examples, please check correctness.
Added Virtual energy service, please compare against assumptions.
Did some cleanup and fixed errors in adjacent sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,11 @@
    * [Battery charge controller service](#battery-charge-controller-service)
    * [Gateway service](#gateway-service)
    * [Version service](#version-service)
-   * [Logging service](#logging-interfaces)
-
+   * [OTA service](#OTA-service)
+   * [Virtual energy service](#Virtual-energy-service)
+   * [Technology specific service](#Technology-specific-service)
+* [Common interfaces](#interfaces)
+   * [Logging interface](#logging-interface)
 
 ## Service overview
 
@@ -1328,22 +1331,6 @@ Name            | Value type   | Description
 `sdk_library`   | string       | SDK Library type (manufacturer internal)
 `protocol`      | string       | Protocol version (Z-Wave)
 
-### Logging interfaces 
-
-#### Service name 
-
-Command can belong to any service of any application or adapter.
-
-#### Interfaces 
-
-Type | Interface             | Value type | Properties | Description 
------|-----------------------|------------|------------|--------------
-in   | cmd.log.set_level     | string     |            | 
-in   | cmd.log.get_level     | null       |            | 
-out  | evt.log.level_report  | string     |            |   
-
-Supported log level : `trace`,`debug`,`info`,`warn`,`error`
-
 ### OTA Service 
 This service is used to manage over-the-air (OTA) upgrades of end devices.
 
@@ -1384,6 +1371,63 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual energy service
+
+The service allows devices with output binary switch service or thermostat service to report accumulated energy consumption in case they do not have their own electric measurement or metering service. Thank to this service delivered energy is calculated based on manually provided power of a device. A device shall calculate energy at every relay state change, mode change or at least every interval value - 30 minutes by default.
+
+#### Service names
+
+`virtual_energy`
+
+#### Interfaces
+
+Type | Interface                | Value type | Description
+-----|--------------------------|------------|------------
+in   | cmd.meter.get_report     | null       | Triggers "evt.meter.report" on a remote device.
+in   | cmd.meter.reset          | null       | Sets energy counter on a remote device to 0.
+out  | evt.meter.report         | float      | Measurement of delivered energy. Default unit should be kWh.
+-|||
+in   | cmd.set_interval         | int        | Interval in minutes for energy recalculation. Overwrites a default value.
+in   | cmd.add_device           | int_map    | Adds Virtual energy service to a selected device. The device shall be reporting energy consumption. Map of integers passed as an argument should provide energy consumption for every mode in watts.
+in   | cmd.remove_device        | null       | Removes Virtual energy service from a selected device. The device shall not be reporting energy consumption.
+-|||
+in   | cmd.mode.get_report      | null       | Triggers "evt.mode.report" on a remote device.
+out  | evt.mode.report          | int        | Number of modes supported by a device. In each mode a device may consume various amount of energy.
+
+#### Service props
+
+Name             | Supported values               | Description
+-----------------|--------------------------------|-------------
+`unit`           | "Wh", "kWh", "MWh"             | If energy delivered is less then 1 kWh, unit should be Wh
+
+#### Examples
+
+Adding Virtual energy service on a device working in a one of three available modes (eg. 0-standby, 1-heating, 2-fan):
+
+```json
+{
+   "type": "cmd.add_device",
+   "serv": "virtual_energy",
+   "val_t": "int_map",
+   "val": {
+      "0": 10,
+      "1": 1500,
+      "2": 250
+   }
+}
+```
+
+Virtual energy service measurement reporting energy delivered value equal 125.5 kWh:
+
+```json
+{
+   "type": "evt.meter.report",
+   "serv": "virtual_energy",
+   "val_t": "float",
+   "val": { 123.5 }
+   "prop": {"unit": "kWh"}
+}
+```
 
 ### Technology specific service
 
@@ -1407,6 +1451,23 @@ Name        | Value example                 | Description
 ------------|-------------------------------|-------------
 `sub_value` | "home_security:sensor_status" | With usage of sub_value, Vinculum will know that it has to store separate value for every pair of choosen properties for given notification.
 
+## Common interfaces
+
+### Logging interface 
+
+#### Service name 
+
+Command can belong to any service of any application or adapter.
+
+#### Interfaces 
+
+Type | Interface             | Value type | Properties | Description 
+-----|-----------------------|------------|------------|--------------
+in   | cmd.log.set_level     | string     |            | 
+in   | cmd.log.get_level     | null       |            | 
+out  | evt.log.level_report  | string     |            |   
+
+Supported log level : `trace`,`debug`,`info`,`warn`,`error`
 
 #### Examples
 
--- a/README.md
+++ b/README.md
@@ -1433,7 +1433,7 @@ Adding Virtual energy service on a device working in a one of three available mo
 }
 ```
 
-Virtual energy service measurement reporting energy delivered value equal 125.5 kWh:
+Virtual energy service measurement reporting energy delivered value equal 123.5 kWh:
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -329,8 +329,7 @@ Name         | Value type                  | Description
 -------------|-----------------------------|-------------
 `delta_t`    | int [seconds]               | Time between the current and the previous meter reading.
 `prv_data`   | float                       | Previous meter reading.
-`unit`       | Defined by prop `sup_units` | Unit of a value in `val` field.
-`prv_unit`   | Defined by prop `sup_units` | Unit of a value in `prv_data` prop.
+`unit`       | Defined by prop `sup_units` | Unit of a value in `val` field and 'prv_data' props field.
 `direction`  | string                      | Defines whether reading is based on consumption ("import") or production ("export").
 `virtual`    | null                        | Informs that energy measurement was calculated by Virtual energy service. 
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@
    * [Inverter services](#inverter-services)
    * [Battery charge controller service](#battery-charge-controller-service)
    * [Gateway service](#gateway-service)
-   * [Version service](#version-service)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
-`meter_elec`    | kWh, kVAh, W, pulse_c, V, A, power_factor | Electric meter
-`meter_gas`     | cub_m, cub_f, pulse_c                     | Gas meter
-`meter_water`   | cub_m, cub_f, gallon, pulse_c             | Water meter
-`meter_heating` | kWh                                       | Heating meter
-`meter_cooling` | kWh                                       | Cooling meter
+Service name    | Measurements                                                                     | Description
+----------------|----------------------------------------------------------------------------------|------------
+`meter_elec`    | energy, power, reactive power, frequency, voltage, current, power_factor, pulses | Electric meters
+`meter_gas`     | volume, energy, pulses, flow, pressure                                           | Gas meter
+`meter_water`   | volume, energy, pulses, flow, pressure                                           | Water meter
```
boguslaw-wojcik, line 313: Why did you remove example units?

README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
 #### Interfaces
 
-Type | Interface                     | Value type | Properties                         | Description
------|-------------------------------|------------|------------------------------------|-------------
-in   | cmd.meter.get_report          | string     | direction                          | Value - is a unit. May not be supported by all meters.
-in   | cmd.meter.reset               | null       |                                    | Resets all historical readings.
-out  | evt.meter.report              | float      | unit, prv_data, delta_t, direction |
-out  | evt.meter_ext.report          | float_map  |                                    | [Extended meter report](#extended-report-object) with up to 17 data points
-in   | cmd.meter_ext.get_report      | null       |                                    | Request extended report
+Dir  | `type`                        | `val_t`            | `prop`                             | Description
```
boguslaw-wojcik, line 319: Why did you change column names?

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual energy service
```
boguslaw-wojcik, line 1402: This is not a device service, this is an adapter service. As such it should be in the adapter documentation, not this page.

README.md
```diff
@@ -345,47 +350,73 @@ Name        | Value example | Description
 Name                | Value example                                  | Description
 --------------------|------------------------------------------------|-------------
 `sup_units`         | ["W", "kWh", "A", "V"]                         | List of supported units.
-`sup_directions`    | ["import", "export"]                           | List of supported directions.
-`sup_extended_vals` | [See extended report](#extended-report-object) |
+`sup_directions`    | ["import", "export"]    (this is not used anywhere? remove?)                       | List of supported directions.
```
boguslaw-wojcik, line 353: Comments should be removed.

README.md
```diff
@@ -345,47 +350,73 @@ Name        | Value example | Description
...
+`sup_extended_vals` | [See extended report](#extended-report-object)  (this is not used anywhere? remove?) |
```
boguslaw-wojcik, line 354: It is used. Remove comment.

README.md
boguslaw-wojcik: Unless your idea is to include this service into any thing that supports virtual energy, which just indicates it is configurable. Once you configure it, one way or another, you will resend the inclusion report?

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+#### Interfaces
+
+Type | Interface                | Value type | Description
+-----|--------------------------|------------|------------
+in   | cmd.set_interval         | int        | Interval in minutes for energy recalculation. Overwrites a default value.
+in   | cmd.add_device           | int_map    | Adds Virtual energy service to a selected device. The device shall be reporting energy consumption. Map of integers passed as an argument should provide energy consumption for every mode in watts.
+in   | cmd.remove_device        | null       | Removes Virtual energy service from a selected device. The device shall not be reporting energy consumption.
+-|||
+in   | cmd.mode.get_report      | null       | Triggers "evt.mode.report" on a remote device.
+out  | evt.mode.report          | int        | Number of modes supported by a device. In each mode a device may consume various amount of energy.
```
boguslaw-wojcik, line 1419: Actually, if you implement it as a device service, you don't need those commands. Just add a service property called supported modes. This is something known at inclusion, it won't change in the lifecycle of the device, so we don't need commands for it.

boguslaw-wojcik: Also, modes should be reported as an array of strings, not just a number of them. How someone will know which index goes to which mode?

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+#### Service names
+
+`virtual_energy`
+
+#### Interfaces
+
+Type | Interface                | Value type | Description
+-----|--------------------------|------------|------------
+in   | cmd.set_interval         | int        | Interval in minutes for energy recalculation. Overwrites a default value.
+in   | cmd.add_device           | int_map    | Adds Virtual energy service to a selected device. The device shall be reporting energy consumption. Map of integers passed as an argument should provide energy consumption for every mode in watts.
```
boguslaw-wojcik, line 1415: Integer map of course makes sense, but if I was to nitpick I would say that it should be a float map. Power and energy readings are reported as floats, so I see no reason why shouldn't we allow floats here as well even if it's a bit of an overkill. Also keys should not be string numbers, but mode names.

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual energy service
+
+The service allows devices with output binary switch service or thermostat service to report accumulated energy consumption in case they do not have their own electric measurement or metering service. Thank to this service delivered energy is calculated based on manually provided power of a device. A device shall calculate energy at every relay state change, mode change or at least every interval value - 30 minutes by default. When service is added on a device, service meter_elec is created. service to send measurements.
```
boguslaw-wojcik, line 1404: `When service is added on a device, service meter_elec is created. service to send measurements.`

Remove last left-over. Maybe say something about re-sending inclusion report?

boguslaw-wojcik (changes requested):
See comments.

README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
------|-------------------------------|------------|------------------------------------|-------------
-in   | cmd.meter.get_report          | string     | direction                          | Value - is a unit. May not be supported by all meters.
-in   | cmd.meter.reset               | null       |                                    | Resets all historical readings.
-out  | evt.meter.report              | float      | unit, prv_data, delta_t, direction |
-out  | evt.meter_ext.report          | float_map  |                                    | [Extended meter report](#extended-report-object) with up to 17 data points
-in   | cmd.meter_ext.get_report      | null       |                                    | Request extended report
+Dir  | `type`                        | `val_t`            | `prop`                             | Description
+-----|-------------------------------|--------------------|------------------------------------|-------------
+in   | cmd.meter.get_report          | string             | direction                          | Triggers evt.meter.report. An argument is "export" or "import".
```
eternalamenity, line 321: This is actually not true.
If by an "argument" You mean a "val", then an argument is a unit.

README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
 
 #### Interface props
 
-Name         | Value example | Description
--------------|---------------|-------------
-`delta_t`    |               | Elapsed time in seconds between the 'Meter Value' and the 'Previous Meter Value' measurements.
-`prv_data`   |               | Previous meter reading.
-`unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
-`direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
+Name         | Value type                  | Description
```
eternalamenity, line 329: Another PR for that kind of changes, please :)
It seems like "Value type" as You described it, is something like a description of an interface props(?)
So it's exactly the same as if You had it all in "Description" column(?)

First row could be a nice example:
You propose:
`delta_t` | int [seconds] | Time between the current and the previous meter reading.

instead of
`delta_t` | 60 | Elapsed time in seconds between the 'Meter Value' and the 'Previous Meter Value' measurements.

README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
+-------------|-----------------------------|-------------
+`delta_t`    | int [seconds]               | Time between the current and the previous meter reading.
+`prv_data`   | float                       | Previous meter reading.
+`unit`       | Defined by prop `sup_units` | Unit of a value in `val` field.
+`prv_unit`   | Defined by prop `sup_units` | Unit of a value in `prv_data` prop.
```
eternalamenity, line 334: Does it make sense to show previous meter reading from different unit than what we receive?
Imo, one always want to send previous meter reading with the same unit as actual event has.

README.md
eternalamenity: sup_directions is also used.

README.md
```diff
@@ -345,47 +350,73 @@ Name        | Value example | Description
...
+#### Examples
+
+Meter reading reported by a energy meter:
+```json
+{
+   "type": "evt.meter.report",
+   "serv": "meter_elec",
+   "val_t": "float",
+   "val": { 123.5 },
+   "prop": {"unit": "kWh", "delta_t": 3600},
```
eternalamenity, line 403: props*

README.md
```diff
@@ -345,47 +350,73 @@ Name        | Value example | Description
...
+```json
+{
+   "type": "evt.meter_ext.report",
+   "serv": "meter_elec",
+   "val_t": "float_map",
+   "val": { 
+    "u1": 237,
+    "i1": 0.12,
+    "p_import": 28.4,
+    },
```
eternalamenity, line 417: imo You shall put props in the example, but make it empty(?)

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+{
+   "type": "cmd.add_device",
+   "serv": "virtual_energy",
+   "val_t": "int_map",
+   "val": {
+      "0": 10,
+      "1": 1500,
+      "2": 250
+   }
+}
```
eternalamenity, line 1435: Could You explain me, when would that message be send?
Who will be a receiver and a sender of this message?

eternalamenity (changes requested):

README.md
boguslaw-wojcik: When user configures virtual energy for a device. But modes should be referred by names not integers.

README.md
```diff
@@ -303,32 +306,34 @@ Meters report consumption over the service.
...
-`unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
-`direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
+Name         | Value type                  | Description
+-------------|-----------------------------|-------------
+`delta_t`    | int [seconds]               | Time between the current and the previous meter reading.
+`prv_data`   | float                       | Previous meter reading.
+`unit`       | Defined by prop `sup_units` | Unit of a value in `val` field.
+`prv_unit`   | Defined by prop `sup_units` | Unit of a value in `prv_data` prop.
+`direction`  | string                      | Defines whether reading is based on consumption ("import") or production ("export").
+`virtual`    | null                        | Informs that energy measurement was calculated by Virtual energy service.
```
boguslaw-wojcik, line 336: Guys, important thing. Properties ARE ALWAYS strings. It's always map of string keys and string values. Even if it is integer it should be returned as "11" for instance. @BudzikM Please do not change the layout of documentation that deviates from existing ones. The column should contain example values.

Also, i'd make virtual value to be "true".

README.md
BudzikM: ok

README.md
BudzikM: yes, it will send inclusion report. The app needs to know that device support virtual energy to enable panel

README.md
eternalamenity: We agreed with @BudzikM that we would like to describe "prv_data" that it's unit is defined in "unit" property.

README.md
```diff
@@ -1384,6 +1399,53 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+Virtual energy service measurement reporting energy delivered value equal 123.5 kWh:
+
+```json
+{
+   "type": "evt.meter.report",
+   "serv": "meter_elec",
+   "val_t": "float",
+   "val": { 123.5 }
+   "prop": {"unit": "kWh", "virtual"}
```
eternalamenity, line 1446: "props": {"unit": "kWh", "virtual" : "true"}

README.md
BudzikM: exemplary json can be added belowe. Feel free to do it.

README.md
BudzikM: @boguslaw-wojcik ? Is props field mandatory, even if empty ? We should state it somwhere if true.

PR #60, opened 2022-07-20
Add virtual energy service

BudzikM:
I extracted this PR from previous PR https://github.com/futurehomeno/fimp-api/pull/59. Since this service is closely connected to elec_meter (and possibly other XXX_meter) let's call it virtual_meter. It will make it more generic. Let it be just a virtual_meter and whatever it is under the masc - electrical device or water flow meter with on/off valve - it can calculate accumulated consumption of every medium reporting it by  XXX_meter service, depends on a typeof device. We should keep FIMP protocol generic imo and possible flexible to be future proved. It will be an extension for any meter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1384,6 +1384,50 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual energy service
+
+The service allows devices with output binary switch service or thermostat service to report accumulated energy consumption in case they do not have their own electric measurement or metering service. Thank to this service delivered energy is calculated based on manually provided power of a device. A device shall calculate energy at every relay state change, mode change or at least every interval value - 30 minutes by default. When service is added on a device, service meter_elec is created. service to send measurements.
+
+#### Service names
+
+`virtual_energy`
+
+#### Interfaces
+
+Type | Interface                | Value type |   Unit  | Description
+-----|--------------------------|------------|---------|------------
+in   | cmd.set_interval         | int        | minutes | Interval  for energy recalculation. Overwrites a default value.
+in   | cmd.add                  | int_map    | watts   | Adds Virtual energy service to a selected device to report energy consumption. Map of integers shall provide power use for every mode.
+in   | cmd.remove_device        | null       |         | Removes Virtual energy service from a selected device. The device shall not be reporting energy consumption.
+
+#### Examples
+
+Adding Virtual energy service on a device working in a one of three available modes (eg. 0-standby, 1-heating, 2-fan):
+
+```json
+{
+   "type": "cmd.add_device",
+   "serv": "virtual_energy",
+   "val_t": "int_map",
+   "val": {
+      "off": 10,
+      "heat": 1500,
+      "fan": 250
+   }
+}
+```
+
+Virtual energy service measurement reporting energy delivered value equal 123.5 kWh:
+
+```json
+{
+   "type": "evt.meter.report",
+   "serv": "meter_elec",
+   "val_t": "float",
+   "val": { 123.5 },
+   "props": {"unit": "kWh", "virtual" : "true"}
+}
+```
 
 ### Technology specific service
 
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@
    * [Battery charge controller service](#battery-charge-controller-service)
    * [Gateway service](#gateway-service)
    * [Version service](#version-service)
+   * [Virtual meter service](#virtual-meter-service)
    * [Logging service](#logging-interfaces)
 
 
@@ -329,7 +330,7 @@ Name         | Value example | Description
 `prv_data`   |               | Previous meter reading.
 `unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
 `direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
-`virtual`    | "true"        | "true" informs that energy measurement was calculated by Virtual energy service. "false" has the same meaning as if the filed is not present.
+`virtual`    | "elec"        | If present, then the measurement was calculated by a virtual service. Allowed options: "elec", "gas", "water", "heating", "cooling".
 
 #### Important notes
 
@@ -1385,30 +1386,30 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
-### Virtual energy service
+### Virtual meter service
 
 The service allows devices with output binary switch service or thermostat service to report accumulated energy consumption in case they do not have their own electric measurement or metering service. Thank to this service delivered energy is calculated based on manually provided power of a device. A device shall calculate energy at every relay state change, mode change or at least every interval value - 30 minutes by default. When service is added on a device, service meter_elec is created. service to send measurements.
 
 #### Service names
 
-`virtual_energy`
+`meter_virtual`
 
 #### Interfaces
 
 Type | Interface                | Value type |   Unit  | Description
 -----|--------------------------|------------|---------|------------
 in   | cmd.set_interval         | int        | minutes | Interval  for energy recalculation. Overwrites a default value.
-in   | cmd.add                  | int_map    | watts   | Adds Virtual energy service to a selected device to report energy consumption. Map of integers shall provide power use for every mode.
-in   | cmd.remove_device        | null       |         | Removes Virtual energy service from a selected device. The device shall not be reporting energy consumption.
+in   | cmd.add                  | int_map    | watts   | Adds Virtual meter service to a selected device to report energy consumption. Map of integers shall provide power use for every mode.
+in   | cmd.remove               | null       |         | Removes Virtual meter service from a selected device. The device shall not be reporting energy consumption.
 
 #### Examples
 
-Adding Virtual energy service on a device working in a one of three available modes "off", "heat" or "fan":
+Adding Virtual meter service on a device working in a one of three available modes "off", "heat" or "fan":
 
 ```json
 {
    "type": "cmd.add_device",
-   "serv": "virtual_energy",
+   "serv": "meter_virtual",
    "val_t": "int_map",
    "val": {
       "off": 10,
@@ -1418,7 +1419,7 @@ Adding Virtual energy service on a device working in a one of three available mo
 }
 ```
 
-Virtual energy service measurement reporting energy delivered value equal 123.5 kWh:
+Virtual meter service measurement reporting energy delivered value equal 123.5 kWh:
 
 ```json
 {
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Name         | Value example | Description
 `prv_data`   |               | Previous meter reading.
 `unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
 `direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
-`virtual`    | "elec"        | Field is presend if the measurement was calculated by a virtual service. Allowed options: "elec", "gas", "water", "heating", "cooling".
+`virtual`    | "true"        | Field is present and equal "true' if the measurement was calculated by a virtual service. "false" or not being present have equal meaning.
 
 #### Important notes
 
--- a/README.md
+++ b/README.md
@@ -330,7 +330,7 @@ Name         | Value example | Description
 `prv_data`   |               | Previous meter reading.
 `unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
 `direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
-`virtual`    | "elec"        | If present, then the measurement was calculated by a virtual service. Allowed options: "elec", "gas", "water", "heating", "cooling".
+`virtual`    | "elec"        | Field is presend if the measurement was calculated by a virtual service. Allowed options: "elec", "gas", "water", "heating", "cooling".
 
 #### Important notes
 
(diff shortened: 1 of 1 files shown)


BudzikM:
[boguslaw-wojcik] Maybe say something about re-sending inclusion report?

[BudzikM] I could, but I thought it is not related to FIMP protocol, is it ?

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual meter service
+
+This service enables a device to report accumulated consumption in case it does not have its own metering capabilities (eg. a relay with output binary switch service or thermostat). Thanks to the service accumulated consumption is calculated based on manually provided consumption in a unit of time (eg. watts, m3/h). A device shall calculate cumulated consumption at every state's change, mode change and at least every interval value - 30 minutes by default. When this service is added on a device, appropriate metering service is created (eg. meter_elec, meter_water) and a device starts to send measurements.
```
Hexagonale, line 1391: Thanks to `this` service, accumulated consumption is...

`The service` shall calculate...

and `the service` starts to send measurements

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+#### Service names
+
+`meter_virtual`
+
+#### Interfaces
+
+Type | Interface                | Value type |   Unit  | Description
+-----|--------------------------|------------|---------|------------
+in   | cmd.set_interval         | int        | minutes | Interval  for accumulated consumption recalculation. Overwrites a default value.
```
Hexagonale, line 1401: overwrites the default, `30 minutes`, value.

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+in   | cmd.add                  | int_float  | define by `unit` props   | Adds Virtual meter service to a selected device to report accumulated consumption. Map of floats shall provide consumption for every mode.
```
Hexagonale, line 1402: Isn't the `value_type` for map of floats something like `float_map`?

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+in   | cmd.remove               | null       |         | Removes Virtual meter service from a selected device. The device shall not be reporting accumulated consumption.
+
+  #### Interface props
+
+Name         | Value example | Description
+-------------|---------------|-------------
+`unit`       | "W", "m3/h"   |  Consumption
```
Hexagonale, line 1409: Consumption `of energy in a unit of time`

Hexagonale (changes requested):
Nits

README.md
boguslaw-wojcik, line 1401: @BudzikM Every single interface we implement has three segments and two dots:
* cmd/evt
* service name or logical component of the service
* action prefixed if applicable with set and get and suffixed if applicable with report

You're not following this convention. Interfaces should be like this:
* cmd.(virtual_)meter.set_interval
* cmd.(virtual_)meter.add(_consumption)
* cmd.(virtual_)meter.remove(_consumption)

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+`meter_virtual`
+
+#### Interfaces
+
+Type | Interface                | Value type |   Unit  | Description
+-----|--------------------------|------------|---------|------------
+in   | cmd.set_interval         | int        | minutes | Interval  for accumulated consumption recalculation. Overwrites a default value.
+in   | cmd.add                  | int_float  | define by `unit` props   | Adds Virtual meter service to a selected device to report accumulated consumption. Map of floats shall provide consumption for every mode.
+in   | cmd.remove               | null       |         | Removes Virtual meter service from a selected device. The device shall not be reporting accumulated consumption.
```
boguslaw-wojcik, line 1403: Command to remove acts like "remove all". Thing if you want to have fine-grained behavior where you can remove consumption set for a single unit and remove all at once. I'd suggest having two interfaces, one for removing a particular unit and one for removing all.

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+This service enables a device to report accumulated consumption in case it does not have its own metering capabilities (eg. a relay with output binary switch service or thermostat). Thanks to the service accumulated consumption is calculated based on manually provided consumption in a unit of time (eg. watts, m3/h). A device shall calculate cumulated consumption at every state's change, mode change and at least every interval value - 30 minutes by default. When this service is added on a device, appropriate metering service is created (eg. meter_elec, meter_water) and a device starts to send measurements.
+
+#### Service names
+
+`meter_virtual`
+
+#### Interfaces
+
+Type | Interface                | Value type |   Unit  | Description
```
boguslaw-wojcik, line 1399: Once again you're changing the documentation conventions, but introducing different columns to the table. If interfaces use properties then there should be a column called property describing properties used by the interface.

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+-------------|---------------|-------------
+`unit`       | "W", "m3/h"   |  Consumption
+  
+#### Examples
+
+Adding Virtual meter service on a device working in a one of three available modes "off", "heat" or "fan":
+
+```json
+{
+   "type": "cmd.add_device",
```
boguslaw-wojcik, line 1417: Your example does not follow the current interface description.

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+   "val_t": "int_float",
+   "val": {
+      "off": 10,
+      "heat": 1500,
+      "fan": 250
+   }
+}
+```
+
+Virtual meter service measurement reporting energy delivered value equal 123.5 kWh:
```
boguslaw-wojcik, line 1428: Not sure if this example is important, because it will be reported BY the meter elec service not virtual meter service.

README.md
boguslaw-wojcik, line 1402: There is no such type as int float.

README.md
```diff
@@ -1384,6 +1386,56 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+in   | cmd.set_interval         | int        | minutes | Interval  for accumulated consumption recalculation. Overwrites a default value.
+in   | cmd.add                  | int_float  | define by `unit` props   | Adds Virtual meter service to a selected device to report accumulated consumption. Map of floats shall provide consumption for every mode.
+in   | cmd.remove               | null       |         | Removes Virtual meter service from a selected device. The device shall not be reporting accumulated consumption.
+
+  #### Interface props
+
+Name         | Value example | Description
+-------------|---------------|-------------
+`unit`       | "W", "m3/h"   |  Consumption
+
```
boguslaw-wojcik, line 1410: You should add section Service props with sup_modes and sup_units to list modes for which unit consumption can be configured.

README.md
```diff
@@ -329,6 +330,7 @@ Name         | Value example | Description
 `prv_data`   |               | Previous meter reading.
 `unit`       | "kWh"         | One of sup_units. For meter_unknown it will be a number.
 `direction`  | "export"      | Defines whether reading is based on consumption (import) or production (export). Applicable for every meter except for meter_elec.
+`virtual`    | "true"        | Field is present and equal "true' if the measurement was calculated by a virtual service. "false" or not being present have equal meaning.
```
boguslaw-wojcik, line 333: "true" instead of "true'

boguslaw-wojcik (changes requested):
Multiple inconsistencies.

README.md
BudzikM: 1st I think is wrong?
2nd - ok
3rd - ok

README.md
BudzikM: I cross-checked and it's not everywhere the same. Why should we duplicate cmd.SERVICE_NAME.sth since service's name is already in the field before ? I believe every protocol should be succinct wihtout repetitions.

README.md
BudzikM: Good point but I am wondering. Virtual meter is added per MQTT address which denotes "%d_%d" (devID, endpoinID). We do not support more then 1 virtual meter per this address. So isn't it unambiguous if we send cmd.remove for any %d_%d address ?

README.md
BudzikM: eg:
```
"type": "cmd.add",
"serv": "meter_virtual",

**Comment by BudzikM** [comment]
> Multiple inconsistencies.

What ? Is this some general comment?

File: README.md (hunk 9 above)
Comment (BudzikM) [reply README.md:1410]: ok

File: README.md (hunk 8 above)
Comment (BudzikM) [reply README.md:1438]: This is just an exemple, should be just helpful.

File: README.md (hunk 5 above)
Comment (BudzikM) [reply README.md]: @boguslaw-wojcik ?

**Comment by boguslaw-wojcik** [comment]
>

File: README.md (hunk 1 above)
Comment (boguslaw-wojcik) [reply README.md]: Unless requirements changed this service should be capable of spoofing both pulse and cumulative measurements, so in the case of electric energy mobile app in my opinion should deliver:

```
{
   "type": "cmd.meter.add",
   "serv": "meter_virtual",
   "val_t": "float_map",
   "val": {
      "off": 1.5,
      "on": 500,
   },
   "props": {
      "unit": "W"
   }
}
```

But in the result added meter_elec service should be both supporting W and kWh.

File: README.md (hunk 2 above)
Comment (boguslaw-wojcik) [reply README.md]: Two reasons:

1) Convention. All existing documented interfaces consist of three segments. The reason we have conventions in software development is that it allows a reader/consumer/developer to quickly establish patterns and base his future expectations on these patterns, treating them as learning heuristics. If you break those conventions and start using free-form interface names, this will have a detrimental effect on the entire development and user experience and increase the risk of bugs. Unless there is a damn good reason to stray off the path, even if convention is discussable, it's better to have any convention than none at all.

image

2) Service component. The middle segment is not a copy of the name of the service, is the name of a component of the service or a logical part of it or a sub-domain of it. Just take the first service from this specification. You see there a config domain and others.

Please add to the interfaces the middle segment like `meter`.

File: README.md (hunk 5 above)
Comment (boguslaw-wojcik) [reply README.md]: @BudzikM Negative, this is not how this behavior should in my opinion be done.

Imagine that you support for a virtual electricity three units: W, A, V. 
* User configures wattage consumption and sends add command for W unit.
* Adapter adds for the first time meter_elec service and declares it supports W and kWh in relevant attributes, then sends updated inclusion report.
* User configures amperage consumption and sends add command for A unit.
* Adapter adds to the already created meter_elec service another supported unit A and sends updated inclusion report.
* User configures voltage and sends add command for V unit.
* Adapter adds to the already created meter_elec service another supported unit V and sends updated inclusion report.

Conclusions:
* You should not create multiple meter_elec services
* Adapter must have a logic based on units it allows to virtualize, understand if they are pulse or cumulative, how to spoof them and calculate, and to which service they should be added.

As for the implementation, currently I think you may just implement it as supporting just one unit (W) and do not hassle about all the complexities, but the design of API should allow such extension in future.

File: README.md (hunk 1 above)
Comment (BudzikM) [reply README.md]: You probably forgot about sth, json you pasted is identical with my propasal

File: README.md (hunk 2 above)
Comment (BudzikM) [reply README.md]: ok.

File: README.md (hunk 5 above)
Comment (BudzikM) [reply README.md]: This is some misconoception partially, we should not base any action on unit. Unit is a unit. It depends on a device what capabilities it has and unit just need to match it. V doesn make any phisical sense, what would be the result of voltage * delta_t ?. Mutliple meters will not be supported on the same device, action is to overwrite the previous one. If a device is kind of meter elec and it supports energy metering, sending any unit different that Watts should return error and device makes this decision.

File: README.md (hunk 8 above)
Comment (BudzikM) [reply README.md:1438]: I can move it to meter_elec paragraph and put a link here, what do you thinkg is better ?

File: README.md (hunk 5 above)
Comment (boguslaw-wojcik) [reply README.md]: Let's have a call.

File: README.md (hunk 1 above)
Comment (boguslaw-wojcik) [reply README.md]: Let's have a call.

File: README.md [hunk 11]
```diff
@@ -1384,6 +1386,66 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
    }
 }
 ```
+### Virtual meter service
+
+This service enables a device to report accumulated consumption in case it does not have its own metering capabilities (eg. a relay with output binary switch service or thermostat). Thanks to the service accumulated consumption is calculated based on manually provided consumption per unit of time (eg. watts, m3/h). The service shall calculate cumulated consumption at every state's change, mode change and at least every interval value. When this service is added on a device, appropriate metering service is created (eg. meter_elec, meter_water) and the service starts to send measurements.
```
boguslaw-wojcik, line 1391: Mention pulse readings as well.

README.md
```diff
@@ -1384,6 +1386,66 @@ out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade s
...
+
+`meter_virtual`
+
+#### Interfaces
+
+Type | Interface                | Value type |  Props  | Description
+-----|--------------------------|------------|---------|------------
+in   | cmd.meter.set_interval         | int        |         | Interval in minutes for accumulated consumption recalculation. Overwrites a default value 30 minutes.
+in   | cmd.meter.add                  | float_map  | `unit`  | Adds corresponding meter service (eg. meter_elec) to a selected device to report accumulated consumption. Map of floats shall provide consumption for every mode.
+in   | cmd.meter.remove               | null       |         | Removes corresponding meter service from a selected device. The device shall not be reporting accumulated consumption.
```
boguslaw-wojcik, line 1403: Indicate it removes all configured units.

README.md
boguslaw-wojcik: Had discussion in a call. Resolved.

boguslaw-wojcik (approved):

BudzikM:
> Nits

??? @Hexagonale

Hexagonale (approved):

PR #69, opened 2022-08-18
Feature/sdu 1873 handle unknown meter

PaulinaDrozda:


README.md
```diff
@@ -1464,6 +1465,16 @@ Type | Interface               | Value type | Description
 out  | evt.notification.report | object     | Notification reports that are sent by devices.
 out  | evt.sensor.report       | object     | Multilevel sensor reports with unknown sensor type or/and unknown unit.
 out  | evt.meter.report        | object     | Meter reports with unknown meter type or/and unit.
+in   | evt.meter.reset         | null       | Resets all historical readings.
+
+#### Service props
+
+Name               | Value example          | Description
+-------------------|------------------------|-------------
+`meter_type`       | "meter_elec", "6"      | Meter type - described as text when known and number if unknown.
```
eternalamenity, line 1474: Probably leave one example value here and type in description that it can be either a numeric value or translated name.

eternalamenity (dismissed):

README.md
```diff
@@ -1464,6 +1465,16 @@ Type | Interface               | Value type | Description
 out  | evt.notification.report | object     | Notification reports that are sent by devices.
 out  | evt.sensor.report       | object     | Multilevel sensor reports with unknown sensor type or/and unknown unit.
 out  | evt.meter.report        | object     | Meter reports with unknown meter type or/and unit.
+in   | evt.meter.reset         | null       | Resets all historical readings.
```
boguslaw-wojcik, line 1468: Does this interface reset ALL unknown meters or just one random, or two random, or?

README.md
boguslaw-wojcik, line 1474: Not sure if I understand the purpose of these properties. This service can be utilized to channel events/readings from multiple meters, so why it has single meter_type property?

The same goes for supported units, export units or rate types. If you need to put it here, then it should be a complex map of supported units per particular meter type. But do we really need this information? We will have it post factum and since we don't implement interface for active query commands, how should we use this information?

boguslaw-wojcik (changes requested):
Some questions.

README.md
PaulinaDrozda: From what I understand services are created per endpoint and since specific endpoints return only one meter type I didn't see a reason to allow multiple meter types here. 
I don't know yet what kind of information is useful for the interface (so that it can show something or expect some specific info before it receives first report from the device), but if you say it's not useful at all then I'll remove these properties :smiley:

README.md
PaulinaDrozda: From what I see in the documentation it resets all readings from the device

boguslaw-wojcik (approved):

PR #74, opened 2022-09-15
SDU-1893: Added new inclusion

PaulinaDrozda:


eternalamenity (approved):

thing-management.md
```diff
@@ -482,6 +482,63 @@ Message (command):
...
+    "val_t": "null",
+    "props": {},
+    "tags": null,
+    "uid": "123456789"
+}
+```
+
+### Inclusion v2 PIN requested
+
+If included device requests a PIN then the following message will sent:
```
boguslaw-wojcik, line 506: will *be* sent

thing-management.md
```diff
@@ -482,6 +482,63 @@ Message (command):
...
+
+Topic: `pt:j1/mt:evt/rt:ad/rn:zw/ad:1`
+
+Message (command):
```
boguslaw-wojcik, line 510: It's not command it's event.

thing-management.md
PaulinaDrozda: Fixed

boguslaw-wojcik (approved):
One comment: I don't like a lack of some kind of session indicator in this process - if someone quits the process in the middle and restarts, it's not apparent if the pin will be delivered to the correct process. Maybe it's not a problem...

PR #77, opened 2022-10-19
Slightly improved template of the service specification.

boguslaw-wojcik:


services/_example_service.md
```diff
@@ -0,0 +1,87 @@
...
+```json
+{
+  "serv": "example",
+  "type": "evt.domain_a.report",
+  "val_t": "object",
+  "val": {
+    "field_a": 123.45,
+    "field_b": "value_2"
+  }
+}
```
chribru, line 65: I understand that we don't want our examples bloated, but should they maybe include all required fields, such as `ver`, `src`, `ctime` and `uid`? Just to avoid encouraging sending any FIMP messages without them, I mean

chribru (approved):
LGTM, but consider adding required fields to the examples or a note about that they need to be included in actual use.

services/_example_service.md
chribru: hmm, I think it's safer to include them then, better than having an example that's actually not a valid message, I would say

boguslaw-wojcik: Will do it in next PR.

PR #71, opened 2022-08-24
Update barrier_ctrl

BudzikM:
stopped_at was not used by anyone.

PR #80, opened 2022-10-31
Second batch of refactored services.

boguslaw-wojcik:


boguslaw-wojcik:
https://futurehomeno.github.io/fimp-api/#/services/generic/door_lock

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
-### Barrier control service
+# Barrier Control service
 
-The service represent devices like garage door openers, barriers, window protection shades, etc.
+The service represent devices like garage doors, barriers, window protection shades, and similar.
```
eternalamenity, line 3: Maybe we should also point out clearly the difference between this service and out lvl switch service. As device group that might use it is very similar. (?)

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
------------------|--------------------------------|-------------
-`sup_notiftypes` | audio, visual                  | supported notification-types, like siren, flashlight
-`sup_states`     | open, closed, closing, opening | supported states
-`sup_tstates`    | open, closed                    | supported target states
\ No newline at end of file
+## Interfaces
+
+| Type | Interface                | Value type | Properties   | Description                                                                                                                                                                |
+|------|--------------------------|------------|--------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| in   | cmd.notiftype.get_report | null       |              | Requests notification types the device is using while opening/closing door.                                                                                                |
```
BudzikM, line 13: 'a device sends' ?

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
+| in   | cmd.notiftype.set        | bool_map   |              | Sets notification type the device is is using while opening/closing door. Configurable notification types are defined in [`sup_notiftypes`](#service-properties) property. |
```
BudzikM, line 14: 1. doble 'is'.

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
+| out  | evt.notiftype.report     | bool_map   |              | Returns notification types the device is using while opening/closing door.                                                                                                 |
+| -    |                          |            |              |                                                                                                                                                                            |
+| in   | cmd.op.stop              | null       |              | Commence emergency stop of any operation.                                                                                                                                  |
```
BudzikM, line 17: succinct: Stops any operation immediatelly.

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
+| -    |                          |            |              |                                                                                                                                                                            |
+| in   | cmd.state.get_report     | null       |              | Requests the current state of the device.                                                                                                                                  |
+| out  | evt.state.report         | string     | `stopped_at` | Reports the current state of the device, one of values defined in [`sup_states`](#service-properties) property.                                                            |
```
BudzikM, line 20: 'stopped_at' change to 'position' or 'cur_pos'. See discussion at https://github.com/futurehomeno/fimp-api/pull/71

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
+| -    |                          |            |              |                                                                                                                                                                            |
+| in   | cmd.tstate.set           | string     |              | Sets the target state of the device to the one of values defined in [`sup_tstates`](#service-properties) property.                                                         |
+
+## Interface properties
+
+| Name         | Required | Example | Description                                                                                                                    |
+|--------------|----------|---------|--------------------------------------------------------------------------------------------------------------------------------|
+| `stopped_at` | No       | `"30"`  | Position as percentage value at which the barrier stopped on emergency halt, where `1` is near closed while `99` is near open. |
```
BudzikM, line 28: this shoudl be 'cur_pos' or sth similr and it should represent current position of the barrier, we need to control and report to see in the app if window shade is 30 % or what and such report chould be sent every ~5% to inform user about the progress

services/generic/barrier_control.md
```diff
@@ -1,36 +1,94 @@
...
+
+## Service properties
+
+| Name             | Type      | Example                                    | Description                                                                                        |
+|------------------|-----------|--------------------------------------------|----------------------------------------------------------------------------------------------------|
+| `sup_notiftypes` | str_array | `["audio", "visual"]`                      | List of supported notification types. Possible values are: `audio`, `visual`.                      |
+| `sup_states`     | str_array | `["open", "closed", "closing", "opening"]` | List of supported states. Possible values are:  `closed`, `closing`, `stopped`, `opening`, `open`. |
```
BudzikM, line 35: Possible -> Allowed sounds more natural here.

BudzikM: List of supported target states is:

services/generic/battery.md
```diff
@@ -1,44 +1,68 @@
...
-```javascript
+## Definitions
 
-{
-   "lvl": 90 ,
-   "health" : 70,
-   "state" : "charging",
-   "temp_sensor" : 40,
-}
+* `event` represents a battery notification and is one of: `low_battery`, `replace_soon`, `replace_now`, `charging`, `charged`, `charge_soon`, `charge_now`.
```
BudzikM, line 26: 'low_battery' and 'replace_soon' are duplications... can we remove them if not a legacy issue ?

services/generic/battery_charge_control.md
```diff
@@ -1,23 +1,27 @@
...
+| in   | cmd.mode.set             | string     | Sets the charge mode.                                                    |
+| out  | evt.mode.report          | string     | Reports the charging mode.                                               |
 
-#### Service props
+## Service properties
 
-Name            | Value example                                                                  | Description
-----------------|--------------------------------------------------------------------------------|-------------
-`sup_modes`     | idle, charging, discharging                                                    | Supported modes.
+| Name        | Type      | Example                               | Description               |
```
BudzikM, line 25: Maybe Example column should be Supported values ? As usually it is so.. ? Examples are below the table.

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
+| in   | cmd.color.set              | int_map    | Sets values for all color components using a map in which key is a [`color_component`](#definitions) listed in [`sup_components`](#service-properties) property. |
+| out  | evt.color.report           | int_map    | Returns map of color components, where value is the component intensity.                                                                                         |
+| -    |                            |            |                                                                                                                                                                  |
+| in   | cmd.color.start_transition | object     | Starts a single color component transition, see the [`transition`](#definitions) object definition for more details.                                             |
+| in   | cmd.color.stop_transition  | str        | Stop fading/enhancing single color component. val = "red"                                                                                                        |
```
BudzikM, line 18: Stop changing brightness of a single color. <- maybe this way ?

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
------------------|----------------------------------------------------------------------------|-------------
-`sup_components` | ["red", "green", "blue"]                                                   | List of supported color components
-`sup_durations`  | [{"min": 1, "max": 127, "step": 1}, {"min": 180, "max": 7620, "step": 60}] | Supported duration steps for transition change (optional). Min and max always reflects seconds.
+| Name             | Type      | Example                                                                      | Description                                                                                                          |
+|------------------|-----------|------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------|
+| `sup_components` | str_array | `["red", "green", "blue"]`                                                   | List of supported color components, see definition of [`color_component`](#definitions) for list of possible values. |
```
BudzikM, line 24: possible values -> supported values. I would stick to this therm. Possible means that there are more values besides the 'possible' ones ? Supported means that others are not supported.

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
+| `green`         | Yes    | Yes    | 0-255   | Defines intensity of green color.                                                                                              |
+| `blue`          | Yes    | Yes    | 0-255   | Defines intensity of blue color.                                                                                               |
+| `warm_w`        | Yes    | No     | 0-255   | Is intensity of warm white light source. Mix of warm white and cold white intensity forms color temperature.                   |
+| `cold_w`        | Yes    | No     | 0-255   | Is intensity of cold white light source. Mix of warm white and cold white intensity forms color temperature.                   |
+| `temp`          | No     | Yes    | 1-65279 | Is the color temperature in Mired (micro reciprocal degree). Actual color temperature supported by end devices is 2700K-6500K. |
+| `amber`         | Yes    | No     | 0-255   | Defines intensity of amber color.                                                                                              |
```
BudzikM, line 42: This should be supported as well in ZIgbee there is no reason we cannot transform RGB into another color system... Or at least we need list of supported system - another message cmd.color.get_system ...

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
+| `cyan`          | Yes    | No     | 0-255   | Defines intensity of cyan color.                                                                                               |
+| `purple`        | Yes    | No     | 0-255   | Defines intensity of purple color.                                                                                             |
 
-- cold_w - is cold white light source intensity. Value range 0-255.
+* `transition` is an object representing a single color transition change with the following structure:
 
-- Mix of warm white intensity and cold white intensity forms color temperature.
+| Field      | Type   | Example | Description                                         |
```
BudzikM, line 48: Again here Example columns is not much helpful, Supported values would do the job and make spec shorter. 

| Field           | Type     | Supported values | Description                                                          |
|---------------|--------|---------------------|-----------------------------------------------------|
| component  | string | 'color_component' values | Selected color for transition                   |
| transition     | string | `"up", "down"`                    | Direction of the transition                       |
| duration        | int    | `<1-127>`                          | Optional. Duration of the transition in seconds    |

BudzikM: It would need some more work for global change. I may nhelp you with it if you like.

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
+| transition | string | `"up"`  | Direction of the transition, either `up` or `down`. |
+| duration   | int    | `127`   | Optional duration of the transition in seconds.     |
 
-- Duration property within cmd.color.start_transition command is optional and means seconds.
+* `transition_range` is an object with the following structure:
 
-- In the Z-Wave protocol, color component transition duration can only be set between 1 - 127 seconds or 1 - 127 minutes. In order to fully support the Z-Wave protocol, sup_durations property can contain an array of ranges with a specific step.
+| Field | Type | Example | Description                        |
+|-------|------|---------|------------------------------------|
+| min   | int  | `1`     | Minimum settable value in seconds. |
```
BudzikM, line 58: Example is not helpful. Supported values with give more insight: like for min <1,max>,
for max woudl be <min, 127> etc.

services/generic/barrier_control.md
boguslaw-wojcik: Not sure what do you mean.

image

Even out_lvl_switch and out_bin_switch can be used in our app to operate blinds (or in fact garage door just without the fancy icon). It's just barrier control is more specialized service. All I think could be a note like this:

> Please note blinds and similar devices might also be supported in a limited way using more generic `out_bin_switch` and `out_lvl_switch services`. 

Is this what you meant?

services/generic/barrier_control.md
boguslaw-wojcik: It's `stopped_at` because it is only sent when device is in `stopped` state.

services/generic/color_control.md
boguslaw-wojcik: Supported values should be either in "Description" column if there are few of them or are referenced just once or in a designated point in Definitions section if this is a long list or are referenced multiple times. The whole problem was that before Example column was used either as example value or enum listing.

Supported values may of a field might be defined either as protocol constant or service property or otherwise. Description column is more flexible for this. I have spent a lot of time working on this template and I thought this through.

BTW: Your supported values column for duration is incorrect. The supported values range is defined in sup_durations service property. The only thing I would improve is a internal link to sup_components and sup_durations properties in the description column.

services/generic/color_control.md
boguslaw-wojcik: Please read this spec more carefully. Transition range object is part linked from sup_durations READ-ONLY service property. It's adapter who provides you this structure for you as a suer to figure at what are min max values. You comment does not make sense to me here.

services/generic/color_control.md
boguslaw-wojcik: amber/cyan/purple are present in 6-channel lights, so this is effectively not a 24 bit gamut but 48 bit one? If Zigbee protocol supports 6-channel colors then yeah, but otherwise, not really.

services/generic/barrier_control.md
eternalamenity: Yes this is what i mean...
Plus, I'm aware there are also Garage Gates which uses out_lvl_switch (e.g. Fibaro Nice Bidi-Zwave module).
In this case those out_lvl_switch-es are not possition aware. So can only start level - up/down and stop.
Z-Wave specification permits out level switches to be no position aware.

Maybe it just needs to be clearified - or added to out_lvl_switch description(?).

services/generic/barrier_control.md
eternalamenity: Maybe if Zigbee ad is able to use it also within other state's reports, e.g. when device is opening, then we should permit it? But why not use a typical out_lvl_switch then for that kind of functionalitites?

services/generic/battery.md
```diff
@@ -1,44 +1,68 @@
...
--|||
-out  | evt.alarm.report   | str_map    |            | val = {"event": "low_battery", "status": "activ"}
--|||
-in   | cmd.health.get_report| null     |            | Request battery health report
-out  | evt.health.report  | int        |            | Battery health in %
-in   | cmd.sensor.get_report | string  |            | Request battery temperature report in celsius.
-out  | evt.sensor.report  | float      | unit       | Battery temperature in celsius.
--|||
-in   | cmd.battery.get_report | null |              | Get full battery report
-out  | evt.battery.report | object |                | Battery full report
```
eternalamenity, line 22: So You checked that none of other non-adapter services uses those interfaces? :D

services/generic/color_control.md
```diff
@@ -1,45 +1,104 @@
...
-in   | cmd.color.stop_transition  | str        | Stop fading/enhancing single color component. val = "red"
+| Type | Interface                  | Value type | Description                                                                                                                                                      |
+|------|----------------------------|------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| in   | cmd.color.get_report       | null       | Requests a map of color component values.                                                                                                                        |
+| in   | cmd.color.set              | int_map    | Sets values for all color components using a map in which key is a [`color_component`](#definitions) listed in [`sup_components`](#service-properties) property. |
+| out  | evt.color.report           | int_map    | Returns map of color components, where value is the component intensity.                                                                                         |
```
eternalamenity, line 15: Actually when tested on Fibaro RGBW2 Controller, this will send only one color per every frame received by RGBW2 Controller.
To clearify, this RGBW2 Controller sends 3 frames in a row (one by one).
Zwave-ad sends then 3 consecutive evt.color.report's - each of them informing about intensity of different color.

Maybe it does not need to be included in this documentation, as it permits that kind of situations.
Especcialy if storage-sub_value is used along with it.

services/generic/color_control.md
eternalamenity: @BudzikM why cmd.color.get_system? You say it's possible that one particular device could implement both color systems?
Why not having both systems included in "color_component" service props then?

services/generic/door_lock.md
```diff
@@ -1,69 +1,138 @@
...
+| `sup_components`            | str_array | `["is_secured", "door_is_closed"]` | List of supported lock configuration components. Possible values are: `is_secured`, `door_is_closed`, `bolt_is_locked`,  `latch_is_closed`.                                                                                                                                    |
+| `sup_op_types`              | str_array | `["timed", "constant"]`            | List of supported operation types. Possible values are: `timed`, `constant`.                                                                                                                                                                                                   |
+| `sup_modes`                 | str_array | `["unsecured", "secured"]`         | List of supported modes. Possible values are: `secured`, `unsecured`, `unsecured_with_timeout`, `unsecured_for_inside_door_handles`, `unsecured_for_inside_door_handles_with_timeout`, `unsecured_for_outside_door_handles`,`unsecured_for_outside_door_handles_with_timeout`. |
```
eternalamenity, line 46: Add note that for timeout options, timeout can be defined in cmd.lock.set_configuration?

services/generic/fan_control.md
```diff
@@ -1,33 +1,40 @@
-### Fan control service
+# Fan Control Service
 
-The service has to be used to control a fan operational modes, speed and receive state updates.
+The service is used to control a fan operational modes, speed and receive state updates.
```
eternalamenity, line 3: Add note that fan can also be controlled over the Thermostat service, but then it's impossible to set the fan power.

services/generic/scene_control.md
```diff
@@ -1,21 +1,46 @@
...
-in   | cmd.scene.set        | string     | Set scene
-out  | evt.scene.report     | string     | Event is generated whenever scene button is pressed on controller.
+| Type | Interface            | Value type | Description                                                            |
+|------|----------------------|------------|------------------------------------------------------------------------|
+| in   | cmd.scene.get_report | null       | Requests the currently active scene.                                   |
+| in   | cmd.scene.set        | string     | Sets the scene to the one provided in the value.                       |
+| out  | evt.scene.report     | string     | Event is generated whenever scene button is pressed on the controller. |
 
-#### Service props
+> Devices designed only as scene controllers may support only `evt.scene.report` interface and send report whenever a button is pressed.
```
eternalamenity, line 17: "...whenever a physical button is pressed."(?)

eternalamenity (approved):
left a few comments, but generally approved

services/generic/barrier_control.md
boguslaw-wojcik: Rewritten to be more understandable.

services/generic/barrier_control.md
boguslaw-wojcik: Style killer... :P Done.

services/generic/barrier_control.md
boguslaw-wojcik: I'd leave it for another discussion. If zigbee can be position aware in all times I'd suggest to either include dedicated interface for level reporting and leaving "stopped_at" as is. But also renaming it to "cur_pos" and changing it in zw implementation is an option too. Putting it on the post mortem list. Leaving as it is now.

services/generic/barrier_control.md
boguslaw-wojcik: Leaving discussion for post mortem.

services/generic/barrier_control.md
boguslaw-wojcik: Okies, I think we can treat this list as closed.

services/generic/battery.md
boguslaw-wojcik: `replace_soon` is native Z-Wave notification, while `low_battery` is custom one. We need both even though they might be almost the same.

services/generic/battery_charge_control.md
boguslaw-wojcik: Negative. I worked this through. It is always possible to show an example value, but it is not always possible to easily explain a complete range of possible values. We have either description column to touch this subject or definition section to provide in-depth description. While it may sound easy to define numeric range 1-255 or list enum here, it's impossible to define object schema here without linking outside.

services/generic/color_control.md
boguslaw-wojcik: It's not brigthness, it's value of the color component. In case of RGB it's the value of R, G or B, in case of cold_w, warm_w, temp it's not brithness but temperature. It's brithness/lightness only in the case of HSL, which we do not support fully.

boguslaw-wojcik: Rewritten to "Stops transition of the single color component provided in the value.`

Good catch, this one slipped.

services/generic/color_control.md
boguslaw-wojcik: Ok, let's consider it a closed list.

boguslaw-wojcik: The thing is...

A device ***supports*** only a subset of ***possible*** values.

I think possible is a correct word here.

Like ZW device supports only RGB but not temp, cold_w and warm_w.

After thinking this out, I don't think changing is feasible.

services/generic/color_control.md
boguslaw-wojcik: Actually we have a case like this. Markus Philipsh Hue adapter supports at the same time RGB and kind of HSL. There will be separate discussion for it. From my PoV, color component list should be enough to infer which system device supports.

Currently on RGB is supported by the mobile app.
ZW supports only RGB or RGB+ACP
Zigbee supports now only RGB
Philips support RGB and kind of HSL.

This is going to be cleared and worked out in post mortem.

services/generic/color_control.md
boguslaw-wojcik: Added links to props.

services/generic/battery.md
boguslaw-wojcik: Not a single adapter in production advertises these interfaces. If you have any information about it, please share it. I believe these additions were made in preparation for household battery support which never went past early PoC.

services/generic/color_control.md
boguslaw-wojcik: This is a big problem.

services/generic/scene_control.md
boguslaw-wojcik: Ok

services/generic/door_lock.md
boguslaw-wojcik: I think it's already as verbose as needed.

services/generic/color_control.md
boguslaw-wojcik: Discussed, ZW needs to cache color components and send report with all.

eternalamenity (approved):

BudzikM (approved):

PR #82, opened 2022-11-08
SDU-2068 devices reachability new approach

eternalamenity:


technology-adapter-api.md
```diff
@@ -26,6 +26,62 @@ Message:
...
+    "ctime": "2018-11-22T23:14:40+0100",
+    "uid":"76533455876765"
+}
+```
+
+### Definitions
+
+* `status` describes a reachability of a device, well-defined statuses include: `UP` and `DOWN`.
+
+* Command value has to be treated as `node_id` (according to Z-Wave) or `uuid` (according to Zigbee). Above description is also applicable for `address` property within event report value.
```
chribru, line 81: should we maybe have it be a string and not int, so that we can support protocols and adapters having addresses which are not integers?

boguslaw-wojcik: Address should always be a string, 100%.

technology-adapter-api.md
```diff
@@ -26,6 +26,62 @@ Message:
...
+```json
+{
+    "serv": "zigbee",
+    "type": "cmd.network.get_node",
+    "val_t": "int",
+    "val": 80,
+    "props": null,
+    "tags": null,
+    "ctime": "2018-11-22T23:14:40+0100",
+    "uid":"76533455876765"
```
boguslaw-wojcik, line 48: Please include complete message including "topic". Also uuid is string with hyphens, just copy random UUID and use it there in the example.

technology-adapter-api.md
```diff
@@ -26,6 +26,62 @@ Message:
...
+
+Topic: `pt:j1/mt:cmd/rt:ad/rn:zigbee/ad:1`
+
+Message:
+
+```json
+{
+    "serv": "zigbee",
+    "type": "cmd.network.get_node",
+    "val_t": "int",
```
boguslaw-wojcik, line 43: It's address so string. Check out this:
https://github.com/futurehomeno/fimp-api/blob/master/thing-management.md#getting-inclusion-report

On adapter side a THING has only one identification and it is an THINGADDRESS.

technology-adapter-api.md
```diff
@@ -26,6 +26,62 @@ Message:
...
+
+Message:
+
+```json
+{
+    "serv": "zigbee",
+    "type": "evt.network.node_report",
+    "val_t": "object",
+    "val": {
+      "address": 80,
```
boguslaw-wojcik, line 67: The same - string

technology-adapter-api.md
```diff
@@ -26,6 +26,62 @@ Message:
...
+    "props": null,
+    "tags": null,
+    "ctime": "2018-11-22T23:14:40+0100",
+    "uid":"76533455876765"
+}
+```
+
+### Definitions
+
+* `status` describes a reachability of a device, well-defined statuses include: `UP` and `DOWN`.
```
boguslaw-wojcik, line 79: I'd prefer if objects have their definitions, so maybe it's better to put here table with node_status object
with property address and status?

technology-adapter-api.md
boguslaw-wojcik: Also please remember that we will want to have similar support eventually in other adapters, including cloud/custom adapters as well as maybe a future matter adapter?

boguslaw-wojcik (changes requested):
Some comments but mostly what Christer already pointed out,

technology-adapter-api.md
chribru: ah, yes, also to avoid confusion that address is sometimes a string and sometimes another type, for sure

technology-adapter-api.md
chribru: what do you mean topic? it shouldn't be part of the payload too? Or do you mean the response topic?

btw, `uuidgen -r` is a nice command which should be already installed on most Debian/Ubuntu systems

technology-adapter-api.md
eternalamenity: But You mean to also have value like that?
```
"val": {
    "node_status": {
        "address": "80",
        "status": "DOWN"
    }
}
```

boguslaw-wojcik: "Or just a separate table name "node_status" to explain all "val" properties?"

technology-adapter-api.md
eternalamenity: fixed

technology-adapter-api.md
eternalamenity: fixed

technology-adapter-api.md
eternalamenity: fixed

technology-adapter-api.md
eternalamenity: fixed

technology-adapter-api.md
boguslaw-wojcik: Message has topic key as well. It's optional, but it should be there. It's for future compatibility of payloads if put on Kafka or other queues outiside of MQTT or IoT Core.

technology-adapter-api.md
boguslaw-wojcik: Maybe: "Command value is an `address` uniquely identifying a thing. In case of Z-Wave `address` is equal to `node_id` within the network, while in Zigbee it is equal to `uuid`. Other adapters may have their own understanding of the address, such as unique ID in the third party cloud API."

eternalamenity: fixed

boguslaw-wojcik (approved):

technology-adapter-api.md
```diff
@@ -26,6 +26,69 @@ Message:
...
+    "type": "cmd.network.get_node",
+    "val_t": "string",
+    "val": "80",
+    "props": null,
+    "tags": null,
+    "ver": "1",
+    "ctime": "2018-11-22T23:14:40+0100",
+    "uid": "e604e951-7afb-4f96-981b-62e905757686",
+    "topic": "pt:j1/mt:cmd/rt:ad/rn:zigbee/ad:1"
+}
```
eternalamenity, line 51: here

PR #85, opened 2022-12-23
SDU-3021: Updated ota service definition

PaulinaDrozda:


services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
+out  | evt.status.report       | object     | Shows update progress and estimated time left. See [`ota_report_object`](#definitions) for detailed information. |
+in   | cmd.update.install      | string     | Starts an update of the device. Value should be the full path to the update file.                                | 
 
+## Definitions
+
+* `ota_report_object` is an object with the following structure:
+
+| Field          | Type   | Example         | Description                                                                                                     |
+|----------------|--------|-----------------|-----------------------------------------------------------------------------------------------------------------|
+| state          | string | `"IN PROGRESS"` | One of the `state` values.                                                                                      |
```
Brzezoo, line 22: Wouldn't it be better to declare this field as an enum?

services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
+| substate       | string | `"TRANSFERING"` | One of the `substate` values.                                                                                   |       
+| progress       | int    | `10`            | Optional, only available in substate `TRANSFERING`. Progress percentage of the update transfer.                 |
+| remaining_min  | int    | `3`             | Optional, only available in substate `TRANSFERING`. Remaining time until the end of the transfer (minutes).     | 
+| remaining_sec  | int    | `35`            | Optional, only available in substate `TRANSFERING`. Remaining time until the end of the transfer (seconds).     |
+| last_update    | int    | `132351314`     | Optional. Unix timestamp of the last started update process.                                                    |
```
boguslaw-wojcik, line 27: Suggestion of possible improvement:
* What about string type and RFC3339?
* Wouldn't it make sense to rename the field and add: "started_at" and "finished_at".

Each object report represents a certain stage in the process, which:
* might have never started (both started_at and finished_at will be skipped)
* started but did not finish
* started and finished

boguslaw-wojcik: @BudzikM @eternalamenity @PaulinaDrozda @Brzezoo

services/meta/ota.md
boguslaw-wojcik: It's declared an enum, the list of values is in Definition section. Possibly @PaulinaDrozda you could add a link to the definition section, like:

```
One of the `state` values as defined [here](#definitions)
````

File: services/meta/ota.md [hunk 3]
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
-   "serv": "ota",
-   "val_t": "object",
-   "val": {
-      "success": true,
-      "error": ""
+      "remaining_sec": 3,
+      "state": "IN PROGRESS",
+      "substate": "TRANSFERING",
+      "last_update": 12625242636
    }
```
GenixPL, line 48: pls leave the old documentation, for now, as well, as we have to implement it right now and the change will be gradual

GenixPL (changes requested):

GenixPL:
in general
we - the app, have to add support for the current/old `ota` service now. 
It's not the first time in which we are encountering smaller or bigger breaking changes, and imo we need a way to provide a way to easily differentiate them. Right now we have to read all possibilities (current and previous, and previous are not present in the documentation) and add special logic that will define which "version" to choose. One of possibilities that I see it to keep `ota` service as it is right now, and introduce `ota_v2`. that would add clear line between them

services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
+in   | cmd.update.install      | string     | Starts an update of the device. Value should be the full path to the update file.                                | 
 
+## Definitions
+
+* `ota_report_object` is an object with the following structure:
+
+| Field          | Type   | Example         | Description                                                                                                     |
+|----------------|--------|-----------------|-----------------------------------------------------------------------------------------------------------------|
+| state          | string | `"IN PROGRESS"` | One of the `state` values.                                                                                      |
+| substate       | string | `"TRANSFERING"` | One of the `substate` values.                                                                                   |
```
BudzikM, line 23: TRANSFERING -> TRANSFER

services/meta/ota.md
BudzikM: RFC3339 -> yes,

started / finished -> yes

services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
+| state          | string | `"IN PROGRESS"` | One of the `state` values.                                                                                      |
+| substate       | string | `"TRANSFERING"` | One of the `substate` values.                                                                                   |       
+| progress       | int    | `10`            | Optional, only available in substate `TRANSFERING`. Progress percentage of the update transfer.                 |
+| remaining_min  | int    | `3`             | Optional, only available in substate `TRANSFERING`. Remaining time until the end of the transfer (minutes).     | 
+| remaining_sec  | int    | `35`            | Optional, only available in substate `TRANSFERING`. Remaining time until the end of the transfer (seconds).     |
+| last_update    | int    | `132351314`     | Optional. Unix timestamp of the last started update process.                                                    |        
+
+
+* `state`: Current state of the update process, can be one of: `"NOT STARTED"`, `"IN PROGRESS"`, `"FINISHED"`.
```
BudzikM, line 30: NOT_STARTED -> IDLE

services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
...
+* `substate`: Gives more detailed information about the current update phase. Available values depend on the value in `state`:
+  * for `"IN PROGRESS"`: `"STARTED"`, `"TRANSFERING"`, `"REDISCOVERY"`, `"DEVICE REBOOTING"`, `"NEEDS USER ACTION TO START"`.
```
BudzikM, line 32: I would resigne from predefined 'substate' enum to avoid adding more values in the future. Instead of this maybe we can have some string 'info' to print text in a generic way.
We could use it to report human-readable report errors, and substate etc ?

services/meta/ota.md
```diff
@@ -7,34 +7,44 @@ This service is used to manage over-the-air (OTA) upgrades of end devices.
 
 #### Interfaces
 
-Type | Interface               | Value type | Description
------|-------------------------|------------|--------------
-out  | evt.ota_progress.report | int_map    | Shows upgrade progress and estimated time left.
-out  | evt.ota_end.report      | object     | Sent on upgrade end with upgrade status.
+Type | Interface               | Value type | Description                                                                                                      |
+-----|-------------------------|------------|------------------------------------------------------------------------------------------------------------------|
+in   | cmd.status.get_report   | null       | Requests update progress information.                                                                            |
```
BudzikM, line 12: progress -> status ?

services/meta/ota.md
BudzikM: I need more time, I will review all changes soon.

> in general we - the app, have to add support for the current/old `ota` service now. It's not the first time in which we are encountering smaller or bigger breaking changes, and imo we need a way to provide a way to easily differentiate them. Right now we have to read all possibilities (current and previous, and previous are not present in the documentation) and add special logic that will define which "version" to choose. One of possibilities that I see it to keep `ota` service as it is right now, and introduce `ota_v2`. that would add clear line between them

Ota_v2 will have all different topics, so will not be confused with v1.  You want to specify cleary documentation version or you are asking how to check and recognize ota version in practice ? Sending ota.status.get_report might be a way.

boguslaw-wojcik: @BudzikM It's more like @GenixPL is requesting that both old and new interfaces are kept in the documentation in a visible way so at any time in the future frontend can easily check old versions to maintain backwards compatibility while refactoring the app.

* When (or if) we create a FIMP v2 it will be a separate set of docs with everything reworked which is at least 1.5-2 years away from production. So if we create V2, entire current state of documents will be copied and visible in previous version section in a real time. It might even be possible we will be maintaining and updating FIMPv1 and FIMPv2 at the same time for some possibly long period as well.

* We are not introducing breaking changes to the existing V1 version of FIMP with following expectations:
** something was never used in production or supported, then we can simply remove it
** something was ill-implemented and we need to introduce breaking changes to fix it
** something was used internally and we can handle a three-stage update

So...

services/meta/ota.md
boguslaw-wojcik: This is what I suggested in the meeting. From my PoV it can all be squashed into following state:

NOT_STARTED
IN_RPOGRESS
NEEDS_USER_ACTION
FAILED (or FINISHED_WITH_ERRORS)
COMPLETED (or SUCCESSFUL or FINISHED_SUCCESSFULLY or COMPLETED_SUCCESSFULLY)

If we need more info we can add maybe a `reason` field for a reason why something failed, like this `LOW_BATTERY`, `INVALID_FILE`, etc.

Furthermore, all ENUMS should be non-whitespaced string. I see no reason for CAPITAL ENUMS as most other enums in FIMP documentation are simple snake case also.

services/meta/ota.md
boguslaw-wojcik: NOT_STARTED is not idle IMHO, it means never_started, as after first OTA update the adapter should always return FINISHED or COMPLETED.

services/meta/ota.md
boguslaw-wojcik: Please mark it as deprecated in the description column.

services/meta/ota.md
boguslaw-wojcik: @PaulinaDrozda This is what I suggest to do:

Keep the old interfaces in the documentation and mark them as deprecated.

We pledge ourselves to keep them in documentation V1 for the eternity.

Going for a Kitchen Knife now to make a blood pledge. @GenixPL happy?

boguslaw-wojcik:
@PaulinaDrozda @BudzikM

I strongly recommend to scrap all discussions above and implement this:

image

services/meta/ota.md
```diff
@@ -1,18 +1,53 @@
...
+-----|-----------------------------|------------|------------------------------------------------------------------------------------------------------------------|
+out  | evt.ota_progress.report     | int_map    | Shows update transfer progress and estimated time left, see [`ota_progress`](#definitions).                      |
+out  | evt.ota_start.report        | null       | Sent on upgrade start.                                                                                           |
+out  | evt.ota_end.report          | object     | Sent on upgrade end with upgrade status, see [`ota_end`](#definitions).                                          |
+in   | cmd.ota_status.get_report   | null       | Requests update progress information.                                                                            |
+out  | evt.ota_status.report       | object     | Shows update phase, see [`ota_status`](#definitions).                                                            |
+in   | cmd.ota_update.install      | string     | Z-Wave only. Starts an update of the device. Value should be the full path to the update                         |
```
BudzikM, line 17: Starts application of the new image
or
Starts installation of the new image

boguslaw-wojcik (approved):

services/meta/ota.md
```diff
@@ -1,88 +1,114 @@
...
-      "state": "IN PROGRESS",
-      "substate": "TRANSFERING",
-      "last_update": 12625242636
-   }
+  "type": "evt.ota_end.report",
+  "serv": "ota",
+  "val_t": "object",
+  "val": {
+    "success": true
+  },
```
eternalamenity, line 86: I assume we leave it like that, so it could possibly be extendable in future right?

eternalamenity (approved):

services/meta/ota.md
```diff
@@ -1,88 +1,114 @@
...
-out  | evt.ota_end.report          | object     | Sent on upgrade end with upgrade status, see [`ota_end`](#definitions).                                          |
-in   | cmd.ota_status.get_report   | null       | Requests update progress information.                                                                            |
-out  | evt.ota_status.report       | object     | Shows update phase, see [`ota_status`](#definitions).                                                            |
-in   | cmd.ota_update.install      | string     | Z-Wave only. Starts an update of the device. Value should be the full path to the update                         | 
+| Type | Interface               | Value type | Description                                                                                                |
+|------|-------------------------|------------|------------------------------------------------------------------------------------------------------------|
+| in   | cmd.ota_update.start    | string     | Starts an update of the device. Value should be the full path to the firmware image file.                  |
```
BudzikM, line 13: path could be in url format file:/ for local file or http:// if we want to update from the cloud.

https://en.wikipedia.org/wiki/URL

BudzikM (approved):

GenixPL (approved):

services/meta/ota.md
PaulinaDrozda: This is something we can add in later when this will be implemented in the app etc.

PR #88, opened 2023-01-05
SDU-3112: Added service props for config, removed unused  interfaces

PaulinaDrozda:


services/meta/system.md
```diff
@@ -1,17 +1,15 @@
...
 
 `dev_sys`
 
-#### Interfaces
+## Interfaces
 
 Type | Interface                   | Value type | Description
 -----|---------------------------- |------------|------------
 in   | cmd.config.get_report       | str_array  | Requests service to respond with config report. If array is empty - report all parameters.
-in   | cmd.config.get_supp_list    | null       | Requests service to respond with a list of supported configurations.
```
PaulinaDrozda, line 12: Removed since neither Zigbee nor Z-Wave nor the app heard of this

eternalamenity (dismissed):

services/meta/system.md
```diff
@@ -33,7 +31,35 @@ in   | cmd.node_reinterview        | str_map    | Value example `{"max_age":"1"}
...
-#### Notes
+## Service props
+
+| Name                | Type   | Description                                                                                  |
+|---------------------|--------|----------------------------------------------------------------------------------------------|
+| `sup_configuration` | object | Optional. List of supported configuration params, see [`configuration_param`](#definitions). |
+
+## Definitions
+
+* `configuration_param` is an object of the following structure:
```
boguslaw-wojcik, line 42: @PaulinaDrozda Have you discussed this with @BudzikM ? How does it fit Zigbee usage?
Do we need size property? 
Format should be an enum with snake_case - spaces suggest a freeform text.
In case of checkbox, how do we know how many "options" are there to select and how to label them?

boguslaw-wojcik: Really important note. We're not making FIMP API a copy of ZWave or Zigbee spec. We abstract it for both and future specs too. Please keep this in mind.

services/meta/system.md
```diff
@@ -33,7 +31,35 @@ in   | cmd.node_reinterview        | str_map    | Value example `{"max_age":"1"}
...
+| `name`          | string | `"Temperature alarms"` | Name of the parameter.                                          |
+| `format`        | string | `"radio buttons"`      | Format of the parameter. One of [`param_format`](#definitions). |
+| `size`          | int    | `2`                    | Size of the parameter. Can be one of: 1, 2 or 4.                |
+| `min`           | int    | `-3`                   | Minimum value of the parameter.                                 |
+| `max`           | int    | `50`                   | Maximum value of the parameter.                                 |
+| `default`       | int    | `10`                   | Default value of the parameter.                                 |
+| `read_only`     | bool   | `true`                 | Defines whether parameter is non-modifiable.                    |
+| `advanced`      | bool   | `false`                | Defines whether parameter is advanced.                          |
+| `alters_device` | bool   | `false`                | Defines whether parameter will alter device capabilities.       |
```
BudzikM, line 55: I am against adding a service which agregates in a sink all possible configurations which are not added to specific services.

BudzikM: Following this way of thinking we could have just a one service with addr / cmd type / parametr type / value. It woul dbe too much generic. We need to try to provide nice-to-use API wth division to services, some logic units etc..

BudzikM: Its adapter's role to translate parameters, group them to match specific service and consitute a logic part.

services/meta/system.md
BudzikM: I would maybe need some device configuration, I can think of some generic settings like poll interval (wakeup interval) which all Zigbee device has and assumably ZWave ones as well. But we need to name it what it is and what it does so user could in a friendly way check how does it affect the device.

services/generic/parameters.md
```diff
@@ -0,0 +1,56 @@
...
+| label | string | `"Option 1"` | Label of the option. |
+| value | int    | `1`          | Value of the option. | 
+
+* `parameter_value` is an object of the following structure:
+
+| Field        | Type                  | Example | Description                                                                                              |
+|--------------|-----------------------|---------|----------------------------------------------------------------------------------------------------------|
+| parameter_id | string                | `"1"`   | Id of the parameter.                                                                                     |
+| value_type   | string                | `"int"` | Type of the value. Can be either`"int"` (for `input` and `select`) or `"int_array"` (for `multiselect`). |
+| value        | based on value_type   | `1`     | Value of the parameter.                                                                                  |
```
boguslaw-wojcik, line 54: What would you say if we had strongly typed object with fields:
- value - int
- values - int array

@PaulinaDrozda @BudzikM @eternalamenity 

This way, there is no partial parsing and whatnot? Also since we implement ints only at this stage,  if needed we can extend this object further. Personally, I prefer to avoid undefined types at all costs.

PaulinaDrozda: Ooo, way better, I think

PaulinaDrozda: But if we want to have a possibility to extend this further then maybe name the fields in an even more detailed manner, like "int_value" "int_map_value", this way anything more that we add can follow the same principle. What do you think?

boguslaw-wojcik: It's pure aesthetics.

The thing is that if we use int_value here we also need to use int_value in option object in the params specification and int_default. If so, then it's better to already create value object and have it in spec.Default, spec.options.value and command/report as well.

boguslaw-wojcik: Something like this:

Shared object

```
{
"value_type": "int",
"int_value": 1,
...
}
```

Command/Report object:

```
{
"parameter_id": "1",
"value": {
   "value_type": "int",
   "int_value": 1,
}
}
```

And the same in spec:

```
{
...
"default": {"value_type":...}
"options": [
  {
      "label": "x",
      "value": {"value_type"...}
  }
]

}
```

boguslaw-wojcik (dismissed):
@PaulinaDrozda I added following changes and fixes:
- small formatting issues or wording changes
- value in options is also a value object
- IMPORTANT: param report is an array and by default returns all parameters - I can't imagine an app would send 50 commands to get all params when rendering the form in mobile app.

boguslaw-wojcik:
@GenixPL Please have a quick look at the service to make sure it's ok for frontend to implement in undisclosed future a UI for configuration.

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+|---------------|--------|--------------------------------------------------------------------------|--------------------------------------------------------------------------------|
+| parameter_id  | string | `"1"`                                                                    | ID of the parameter.                                                           |
+| name          | string | `"Sensitivity of the PR sensor"`                                         | Name of the parameter.                                                         |
+| description   | string | `""`                                                                     | Description of the parameter.                                                  |
+| widget_type   | string | `"select"`                                                               | Format of the parameter, one of [`widget_type`](#definitions).                 |
+| value_type    | string | `"int"`                                                                  | Type of the value.                                                             |
```
GenixPL, line 36: Screenshot 2023-01-17 at 13 26 07
do we need it to be duplicated?

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+| options       | object | `[{"label":"Option 1", "value": {"value_type": "int", "int_value": 1}}]` | Only for `select` and `multiselect`. Array of [`select_option`](#definitions). |
+| min           | int    | `-1`                                                                     | Only for `input`. Minimum possible value of the parameter.                     |
+| max           | int    | `20`                                                                     | Only for `input`. Maximum possible value of the parameter.                     |
```
GenixPL, line 39: there will probably be a need to have float value with min and max floats

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+| parameter_id  | string | `"1"`                                                                    | ID of the parameter.                                                           |
+| name          | string | `"Sensitivity of the PR sensor"`                                         | Name of the parameter.                                                         |
+| description   | string | `""`                                                                     | Description of the parameter.                                                  |
+| widget_type   | string | `"select"`                                                               | Format of the parameter, one of [`widget_type`](#definitions).                 |
+| value_type    | string | `"int"`                                                                  | Type of the value.                                                             |
+| options       | object | `[{"label":"Option 1", "value": {"value_type": "int", "int_value": 1}}]` | Only for `select` and `multiselect`. Array of [`select_option`](#definitions). |
```
GenixPL, line 37: just making sure - this is intended to be english only advanced feature (when implemented in the app)?

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+| min           | int    | `-1`                                                                     | Only for `input`. Minimum possible value of the parameter.                     |
+| max           | int    | `20`                                                                     | Only for `input`. Maximum possible value of the parameter.                     |
+| default_value | object | `{"value_type": "int", "int_value": 1}`                                  | Default value, see [`value`](#definitions).                                    |
+| read_only     | bool   | `false`                                                                  | If true value cannot be set.                                                   |
+
+* `widget_type` can be one of:
+    * `input` - input a single value,
+    * `select` - select one item from supported options,
+    * `multiselect` - select multiple items from supported options.
```
GenixPL, line 46: 👍🏻

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+
+`parameters`
+
+## Interfaces
+
+| Type | Interface                 | Value type | Description                                                                                                  |
+|------|---------------------------|------------|--------------------------------------------------------------------------------------------------------------|
+| in   | cmd.sup_params.get_report | null       | Requests information about known parameters.                                                                 |
+| out  | evt.sup_params.report     | object     | Reports known parameters of the device, contains array of [`sup_parameter`](#definitions) objects.           |
+| in   | cmd.param.set             | object     | Sets parameter, see [`parameter_value`](#definitions).                                                       |
```
GenixPL, line 15: so we cannot set multiple parameters at once?

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
...
+| in   | cmd.param.get_report      | str_array  | Requests report with currently set values for specified parameters.                                          |
+| out  | evt.param.report          | object     | Reports current value for specified parameters, contains array of [`parameter_value`](#definitions) objects. |
+
+
+## Service properties
+
+| Name      | Type      | Example | Description                                                |
+|-----------|-----------|---------|------------------------------------------------------------|
+| sup_sizes | int_array | `[1,2]` | Optional. Defines supported sizes of the parameter values. |
```
GenixPL, line 24: i dont think i get this one

services/meta/parameters.md
```diff
@@ -0,0 +1,149 @@
+# Parameters Service
+
+Parameters service allows configuration of a device.
+
+## Service name
+
+`parameters`
```
GenixPL, line 7: in general this doc is hard to read due to the "flat" structure - all elements can put stuff on the same lvl (`min`/`max` and `options`), and we end up with optional stuff everywhere.

services/meta/parameters.md
GenixPL, line 37: `{"value_type": "int", "int_value": 1}` and then we will have `float_value` and `string_value` and `bool_value`?

boguslaw-wojcik: If we need it in future, yes. We design for future extendability.

services/meta/parameters.md
boguslaw-wojcik: Good point, though I think you don't have a need for it. As modern UI is usually applying changes on the fly and does not behave like old-school forms. A loose idea @PaulinaDrozda: 

what if we made this bunch of interfaces:

cmd.param.set -> evt.param.report (for single param, and storage per param ID)
cmd.param.get -> evt.param.report (for single param, and storage per param ID)

optionally if adapter allows it:

cmd.params.get -> evt.params.report (get all parameters, and just one storage)

Just a loose idea - maybe we can recap once again...

services/meta/parameters.md
boguslaw-wojcik: Yes. Labels will be provided by the producer.

boguslaw-wojcik (approved):

PR #91, opened 2023-01-11
Removed switch type property and added properties for supported featu…

boguslaw-wojcik:
…res.

### DIFF
--- a/services/generic/output_level_switch.md
+++ b/services/generic/output_level_switch.md
@@ -33,11 +33,12 @@ Output level switch service acts as a level controller for devices such as dimme
 
 ## Service properties
 
-| Name      | Type   | Example    | Description                                                         |
-|-----------|--------|------------|---------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


boguslaw-wojcik:
1) I see not a single ZW/Zigbee device has a property sw_type on_off/up_down. Only just a couple of legacy enocean devices and custom adapters have it. Maybe it should be removed from the specification, as frontend reports no usage of it and it is not widely supported.

@chribru @bjhandeland Do you know any reason we should keep it in docs?

2) Interfaces for set and start have properties that sometimes are and sometimes are not supported. Maybe it is good idea to add service props indicating it?

This is mostly a question to our beloved ZWave tandem @PaulinaDrozda and @eternalamenity , but also it would be nice to get feedback from @BudzikM .

Please consider this PR as a discussion thread, but at the same time please be concise and provide your views with argumentation.

BudzikM (approved):

eternalamenity (approved):

PR #99, opened 2023-03-10
Refactored newtwork management and common interfaces.

boguslaw-wojcik:
* Rewritten adapter network management section.
* Introduced common interfaces recommendations.


adapter/network_management.md
```diff
@@ -0,0 +1,253 @@
...
+| in   | cmd.network.get_node         | string     | Requests state of a thing with the provided address.                                                              |                         
+| out  | evt.network.node_report      | object     | Reports state of a single thing, see [`node_report`](#definitions) object.                                        |
+| in   | cmd.network.get_all_nodes    | null       | Requests state of all things.                                                                                     |                         
+| out  | evt.network.all_nodes_report | object     | Reports state of all things, an array of [`node_report`](#definitions) objects.                                   |
+| in   | cmd.network.update           | string     | Requests update of the network topology, see [`network_update_mode`](#definitions) definition for allowed values. | 
+| in   | cmd.network.node_update      | int        | Requests an interview and discovery of its capabilities for the provided address of a thing.                      |
```
boguslaw-wojcik, line 43: Issue 1: `cmd.network.node_update` interface breaks from accepted fact that thing address is elsewhere referenced as a string. While Z-Wave node ID and Zigbee UDID are numeric and in Cliffhanger I have introduced internal mapping between public numeric addresses and internal freeform text identifiers (like UUID used by a cloud system of a third party), there are still some old adapters using non-numeric identifiers for thing address.

We have two options:
* Turn our eyes away from this inconsistency and pretend we did not notice it
* Specify in the documentation that it is string but pledge eternal support in the zigbee and z-wave adapter for both integer and string input on this interface handler. This interface is already implemented this way in Z-Wave.

I am specifically asking @GenixPL  @BudzikM  @eternalamenity @chribru for input on this issue.

adapter/network_management.md
```diff
@@ -0,0 +1,253 @@
...
+| in   | cmd.network.reset            | null       | Requests forceful removal of all things and resets the network.                                                   |
+| out  | evt.network.reset_done       | null       | Reports that network reset has been completed.                                                                    |
+
+### Definitions
+
+* `node_report` is an object with the following structure:
```
boguslaw-wojcik, line 49: Issue 2: `node_report` object copies verbatim 5 values  from the inclusion report, which is redundant. However, 3 of these values, even though they are stored verbatim in Vinculum DB, are not processed and not provided by Vinculum in thing/device registry reports. As such they are not forwarded by Cloud Bridge to Niflheim/Site Structure. In essence they are unavailable for mobile application nor any application running on the hub.

We could of course wait for Vinculum,CloudBridge and Niflheim to be updated, but due to update constraints it seems unrealistic and impossible to meet the general timetable for signal quality monitoring features.

Comment if you have better ideas.

adapter/network_management.md
```diff
@@ -0,0 +1,253 @@
...
+| comm_tech       | string | `"zw"`                   | Equal to `comm_tech` field in [inclusion_report](/adapter/thing_management.md#definitions) for this thing.                          |
+| status          | string | `"UP"`                   | Status of a thing, either `UP` or `DOWN`, see [`node_status_matrix`](#definitions).                                                 |
+| sub_status      | string | `"online"`               | Sub status of a thing, one of `online`, `sleeping`, `updating`, `configuring`, `offline`, see [`node_status_matrix`](#definitions). |
+| conn_quality    | string | `"high"`                 | Network connection quality rating, one of `low`, `medium`, `high`, or `undefined` if unknown or not applicable.                     |
+| conn_type       | string | `"direct"`               | Either `direct` or `indirect`, if communication is going through hops or third party cloud services/proxies.                        |
+
+* `node_status_matrix` is a matrix of possible `status` and `sub_status` values:
```
boguslaw-wojcik, line 64: Issue 3: This is a proposal of a status/sub_status matrix. Feel free to propose changes or alternatives. The goal is to have in future a single and readily available place to be able to asses status of the thing, it's connection quality and readiness, which may be impacted by temporary process such as discovery, firmware update or sleeping.

What we need to consider are also UX and certification requirements, an application might need to show the user information that the device i sleeping and any issued command might be received and executed later. We will also want to show the type of connection of the device. It is important to make it easier for the mobile app to access and use this data, as well as other services running in background on the hub.

adapter/network_management.md
boguslaw-wojcik: image

adapter/network_management.md
boguslaw-wojcik: image

adapter/network_management.md
GenixPL: when it comes to the app we dont have much problems with playing around the types (it's a messy everywhere so one more place doesnt matter). we would have substantial amount of work, and complicate things, if we were needed to create two different requests based on some version (plus it can lead to bugs), so i would omit that as much as possible

boguslaw-wojcik: The solution I see is:
* Z-Wave and Zigbee current/old versions accept only integers
* Z-Wave and ZIgbee new versions will accept both strings and integers
* Mobile app will be sending integers to Z-Wave/Zigbee to maintain compatibility and string to other adapters (currently no other adapter supports it).

Considering that It might be unnecessary purism.

adapter/network_management.md
eternalamenity: Zwave ad can extend this report with needed values.

adapter/network_management.md
eternalamenity: It seems okay

adapter/network_management.md
eternalamenity: EDIT: But some properties are already available for SDU App: hash, alias?
Could You point out which properties are propagated to site structure and which are not?

boguslaw-wojcik: Yes, but ThingsplexUI utilizes hash and alias from this report. Nothing else is available in site structure.

eternalamenity: I meant, that I can clearly see e.g. alias for the whole lifetime of a device in SDU App->device details section... Isn't that enough? I mean, if it's visible there then app has to have it stored somewhere?

adapter/network_management.md
budzonmichal: Lets switch to int and keep consistent

adapter/network_management.md
BudzikM: UP/DOWN I call it reachability in the code. We have bunch of 'statuses' everywhere so maybe lets use REACHABILITY as more outstanding name

BudzikM: It does not seem good to me.
- Online and offline seems as duplication of UP/DOWN

What if device was being updated and now request started to time out ? We need to separate these and define what they mean.

First we need to define what each category is about and describe separately to ensure they are disjoint - only then this matrix makes sense.
- STATUS defines if the device will response when inquired. I call it REACHABILITY in zigbee-ad to differentiate from many different 'statuses' elsewhere so I propose the same name here. AVAILABILITY or RESPONSIVE is another option.
- SUB_STATUS hard to say for me what would it be as it's a mixture of different things. Device can be updating and sleeping in the meantime jsut after getting the block od image...
This maybe should be separated to "OTA stage" (updating, discovering) and reachability (online, offline, sleeping)

Hmmm maybe **REACHABILITY** could be set {UP, DOWN, SLEEPING} -> would make more sens from my perspective. Exactl definition to be defined but just to share the idea it would be:
**DOWN**  for:
- AC devices when not reachable as they surely should be responsie all time
- DC devices when no responsive wihin interval <wakeUpInterval, inf>

**UP**  for:
- AC, DC devices recently (recetly - TBDefined) responsive

**SLEEP**  for:
- DC devices which were responsive wihtin  <0, wakeUpInterval>


SUB_SATUS -> maybe rename as LifePhase (I use the name to mark device as Associated (when broadcast packet before inclusion caught), Joined/Included, Discovered, Updating, Discovering, Excluding)
-> This is the state during device which device can be when connected to the network and its important how it behaves depending what actually is happening. Eg: I set device LifePhase to Discovery to reject all cmd.meter.get_report untill whole cluster metering is not discovered & divisors downloaded.

adapter/network_management.md
eternalamenity: I think we should stick with string, because of other adapters or playground/integration apps that do not need to use integers as it's addresses, correct @boguslaw-wojcik ?

adapter/network_management.md
boguslaw-wojcik: @BudzikM 
Thanks for your comments. It's exactly what you said just the other way round. Let me elaborate better and then we can focus on better wording.

Status is the most important binary information:

DOWN - means that the device is considered to be offline/disconnected/unreachable/etc. If it is AC or battery FLIR (non-sleeping) device it means it failed to respond to a number of retries recently and did not send anything on its own. If it is a sleeping device it means it failed to send anything past its sleeping interval usually with some safety margin and did not respond to ping or queued commands during the wakeup period.

UP - just invert the description of DOWN.

Let me put additional notes. Device that you're busy interacting with or is conducing some internal process is UP.  If you're discovering/configuring the device, it is UP, even though you cannot effectively use it or send commands to it. If the device is actively downloading OTA image, it is UP, even though it is not useable at the moment. If device is sleeping, but it is expected for it to sleep now, it is also UP. 

For backwards compatibility we cannot change the property name "status" and we cannot change the values "UP" and "DOWN".

SUBSTATUS is a more detailed information indicating not only if the device is online but also if it is ready to receive commands. SUBSTATUS contains in itself STATUS but for backwards compatibility we need to maintain both of them, so effectively SUBSTATUS repeats and extends what STATUS says but also adds more information.

The matrix shows only all logical combinations of status and substatus.

We can rename substatus and its values however we want but it won't change the number of enums or their logical relation. Maybe instead of online/offline ready/not-ready would be better? I don't know. I don't like not-ready because it is too weak semantically. Das ding ist kaput und reportieren "not-ready"???!!!

boguslaw-wojcik: Minor comments:
* DC does not mean battery powered in FIMP, it means DC source, and it was meant mostly to be used for inverters  and solar panels.
* We probably don't need to map 1:1 with internal statuses kept in the adapter, mostly because we're interested in broadcasting a status from the time of first inclusion report being sent to the time when exclusion report is being sent.

So statuses like associated, joined, discovered probably do not make much sense. But it makes sense to indicate "discovering" status when you have enough information to send initial inclusion report, but then still doing gathering information about the device.

eternalamenity: 1. "What if device was being updated and now request started to time out "
Then just use "Online" status once again if it's online. More descriptive info should be within OTA service.

2. "Device can be being updated and sleeping in the meantime just after getting the block of image..." 
You mean, that the image uploading to device has been done, but installation did not proceeded yet?

@boguslaw-wojcik maybe we should point out a proper usage of those sub_statuses? Like.. for substatus "updating", app would know based on that, that it shall listen and act on ota service and show statuses from that service or...?

BudzikM: You want to ban some combination of this 'matrix' ? 
Please give me examples and under what condition its possible of:
-device is UP and offline
-device is DOWN and online
-device is DOWN and sleeping
-device is DOWN and updating

BudzikM: So according to you "SUBSTATUS is a more detailed information" and STATUS is less detailed information about the same then ? It needs to be precisely defined what is what otherwise it would be the source of confusion for FIMP users.

adapter/network_management.md
boguslaw-wojcik: Yes., @eternalamenity

eternalamenity (approved):

PR #102, opened 2023-03-27
Vinculum storage definition extension and update.

boguslaw-wojcik:
* Definition of storage policy and its strategies.
* Update of notation and all services required to use storage policies.
* Mostly backwards compatible changes in meter service concerning extended report.


chribru (approved):
LGTM

PR #112, opened 2023-08-29
FIMP extensions required for ELVA.

boguslaw-wojcik:


GenixPL (dismissed):
seems fine but reading the  diffs is almost impossible

device_services/generic/chargepoint.md
```diff
@@ -8,38 +8,55 @@ Chargepoint service is used to represent EV chargers.
...
-|-----------------|----------|----------|----------------------------------------------------------------------------------------|
-| `charging_mode` | `"slow"` | No       | One of charging modes defined in [`sup_charging_modes`](#service-properties) property. |
+| Name               | Example                       | Required | Description                                                                                      |
+|--------------------|-------------------------------|----------|--------------------------------------------------------------------------------------------------|
+| `charging_mode`    | `"slow"`                      | No       | One of charging modes defined in [`sup_charging_modes`](#service-properties) property.           |
+| `previous_session` | `"2.5"`                       | No       | Reports energy consumed during previous session in `kWh`.                                        |
+| `started_at`       | `"2022-09-01T08:00:00Z02:00"` | No       | Time of current session start in RFC3339 format.                                                 |
```
BudzikM, line 37: Present only if the current session has started. ? or this is for the previous session when finished ?

BudzikM (approved):

device_services/generic/chargepoint.md
boguslaw-wojcik: Yes. It this is first time the charger is connected to the system it can be empty, otherwise there will always be a current session, even if its finished.

device_services/generic/chargepoint.md
```diff
@@ -8,38 +8,55 @@ Chargepoint service is used to represent EV chargers.
...
+| -    |                                 |            |                                                                     |                                                                                                                        |
+| in   | cmd.current_session.set_current | int        |                                                                     | Sets the offered current for the ongoing session in `A`, must be an integer between `6` and a set `max_current` value. |
+| in   | cmd.current_session.get_report  | null       |                                                                     | Requests energy consumed during the current session and optionally additional characteristics.                         |
+| out  | evt.current_session.report      | float      | `previous_session`, `started_at`,  `finished_at`, `offered_current` | Reports energy consumed during the current session in `kWh` and optionally additional characteristics.                 |
```
GenixPL, line 25: if we decided to split, maybe we should put the `previous_session` and `finished_at` properties under `evt_previous_session.report` with `finished_at` property and not overload **current**`_session` info with **previous**`_session` info?

boguslaw-wojcik: `finished_at` is an attribute of the CURRENT SESSION. 

You start charging, there is only started_at.
You stop charging, there is started_at and stoped at. Session is finished but is still the last session, so it's also a current session. 

Use case: A customer plugs his car at 11 PM an goes to bed. Car starts charging at 11 PM and charges until 2 AM when the session finishes because car is fully charged. In the morning the customer looks on mobile app and sees:

Status: Finished
Started at: 11 PM
Finished at: 2 AM

GenixPL (approved):

PR #136, opened 2024-09-17
Allow multiple sensors of the same type

BudzikM:


device_services/generic/numeric_sensor.md
```diff
@@ -84,7 +86,8 @@ The most popular sensors measure temperature, air humidity, or light intensity.
     "sub_value": "C"
```
eternalamenity, line 86: I dont know if it'll work with the subvalue "c", because every index would be replaced with a new one?

boguslaw-wojcik:
This won't work and will break stuff. We need to discuss this.

eternalamenity:
Consider floor/external temp sensor service as a solution @BudzikM

PR #138, opened 2024-12-23
Add default binary state

BudzikM:


BudzikM:
> Please provide context why do we need this feature.

Most of the devices with on/off functionality included our FH Relay have configuration of the state after power cycle. Some of them by default are ON after power cycle which might be inappropriate/dangerous in some uses, i.e.. power blackout may cause the device to turn on.

We may discuss to add in this service `parameters` for future use to handle this behavior for all services.

device_services/generic/output_binary_switch.md
```diff
@@ -8,11 +8,14 @@ Output binary switch service acts as an on/off switch for wall-plugs, relays, si
...
+| Type | Interface                    | Value type | Description                                                                      |
+|------|------------------------------|------------|----------------------------------------------------------------------------------|
+| in   | cmd.binary.get_report        | null       | Requests report of the binary state.                                             |
+| in   | cmd.binary.set               | bool       | Sets the binary state.                                                           |
+| out  | evt.binary.report            | bool       | Reports `true` when the switch is **on** and `false` when the switch is **off**. |
+| in   | cmd.default_state.get_report | null       | Requests report of the default binary state after power up.                      |
+| in   | cmd.default_state.set        | bool       | Sets the binary state default binary state after power up.                       |
+| out  | evt.default_state.report     | bool       | Reports binary state after power up - `true` then **on**, `false` when **off**.  |
```
eternalamenity, line 18: Add it to parameters service @BudzikM

PR #142, opened 2025-07-28
Add meter transformer interfaces

BudzikM:


device_services/generic/meter.md
```diff
@@ -17,17 +17,21 @@ An electricity meter service can represent a stand-alone AMS meter, like a HAN s
...
+| out  | evt.meter_ext.report             | float_map  |                                          | `split`     |             | Returns an extended electricity report. See [electricity measurements](#electricity-measurements) section for more information.                                                    |
+| -    |                                  |            |                                          |             |             |
+| in   | cmd.meter_transformer.get_report | null       |                                          |             |             | Requests a list of transformation ratios for each measurement affected by the transformer.                                                                                                               |
+| out  | evt.meter_transformer.report     | float_map  |                                          | `split`     |             | Returns a map of transformation ratios for each measurement affected by the transformer.                                                                                            |
```
boguslaw-wojcik, line 33: no point to have storage strategy split here, we can assume that report will always contain ALL units which are transformed.

device_services/generic/meter.md
```diff
@@ -132,6 +136,31 @@ An electricity meter service can represent a stand-alone AMS meter, like a HAN s
...
+{
+  "serv": "meter_elec",
+  "type": "cmd.meter_transformer.set",
+  "val_t": "float_map",
+  "val": {
+   "A": 40,
+   "kWh": 1,
+   "W": 40,
+   "V": 1
+  },
```
boguslaw-wojcik, line 151: command doesn;t have split strategy, commands are not sotred

boguslaw-wojcik (changes requested):
Also add a service param like "sup_transformation_units" or similar listing units which can have a transformation ratio.

See other commands and good to go

PR #143, opened 2025-08-27
Update color_control.md with HSV parameters

BudzikM:
Added hue, saturation, and value parameters to color control documentation.


chribru (approved):
LGTM, what do you think @boguslaw-wojcik ?

chribru:
maybe sat is better than saturation? both are fine with me, tho

boguslaw-wojcik (commented):
I remember there was a very lengthy discussion why this had problems in the mobile. I don't remember why. Please ask @kamilszu when he's back.

boguslaw-wojcik:
@kamilszu Please review this and the behavior of the mobile app.

BudzikM:
FH Mobile App supports RGB components only for now.

Adding "hue" and "saturation" to cmd.color.set command is irrelevant fot the app, what matters is evt.color.report, and for smooth transition for ZGA we can send r,g,b + h,s for now.

